PR #5, opened 2016-01-12
extension api

piedshag:
lets discuss it here

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,10 @@ When a feed is being used hypercore will emit `interested` with the correspondin
 
 This is emitted when a feed is no longer being used.
 
+#### `core.use(extension)`
+
+Use an extension to the protocol. See section below.
+
 ## Feeds
 
 Everytime you write a stream of data to hypercore it gets added to an underlying binary feed. Feeds give you more low-level access to the data stored through the following api.
@@ -165,10 +169,6 @@ Property containing the number of blocks this feed has. This is only known after
 #### `var bool = feed.has(index)`
 
 Boolean indicating wheather or not a block has been downloaded. Note that since this method is synchronous you have to wait for the feed to open before calling it.
-
-### TODO
-
-- [ ] implement extension api
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -166,6 +166,10 @@ Property containing the number of blocks this feed has. This is only known after
 
 Boolean indicating wheather or not a block has been downloaded. Note that since this method is synchronous you have to wait for the feed to open before calling it.
 
+### TODO
+
+- [ ] implement extension api
+
 ## License
 
 MIT
--- a/index.js
+++ b/index.js
@@ -62,6 +62,11 @@ Hypercore.prototype.add = function (opts) {
   return feed(this, null, opts)
 }
 
+Hypercore.prototype.use = function (ext) {
+  // WIP
+  return this.swarm._use(ext)
+}
+
 Hypercore.prototype._close = function (link) {
   var id = link.toString('hex')
   delete this._opened[id]
--- a/lib/protocol.js
+++ b/lib/protocol.js
@@ -9,6 +9,7 @@ var messages = require('./messages')
 
 var MAX_BITFIELD = 4 * 1024 * 1024
 var MAX_MESSAGE = 5 * 1024 * 1024
+var EXTENSION_OFFSET = 64
 
 var ENCODINGS = [
   messages.Join,
@@ -182,9 +183,16 @@ function Protocol (opts) {
   this._remote = []
   this._local = ids()
 
+  // extension variables
+  this._extensions = opts.extensions
+  this._available = []
+  this._remoteExtensions = []
+  self._recieved = {}
+
   var handshake = {
     protocol: 'hypercore',
-    id: this.id
+    id: this.id,
+    extensions: opts.extensions
   }
 
   this._generator.write(messages.Handshake.encode(handshake))
@@ -199,10 +207,74 @@ function Protocol (opts) {
 
 util.inherits(Protocol, duplexify)
 
+// extension api
+
+Protocol.prototype.supports = function (id) {
+  var self = this
+  if (!self._handshook) return
+
+  var offset = id - EXTENSION_OFFSET
+  var extension = self._extensions[offset]
+
+  if (self._available.indexOf(extension) > -1) return true
+  return false
+}
+
+Protocol.prototype.send = function (id, buf) {
+  var self = this
+  if (id >= 255) return false
+  if (!self.supports(id)) return false
+
+  if (!Buffer.isBuffer(buf)) buf = new Buffer(buf)
+
+  var msg = new Buffer(buf.length + 1)
+  msg[0] = id
+  buf.copy(msg, 1)
+  this._generator.write(msg)
+
+  return true
+}
+
+Protocol.prototype.receive = function (extension, cb) {
+  var self = this
+  if (self._recieved[extension]) return
+  self._recieved[extension] = cb
+}
+
+Protocol.prototype._route = function (type, data) {
+  var self = this
+  if (!self._handshook) return
+
+  var extension = self._remoteExtensions[type - EXTENSION_OFFSET]
+  var localId = self._extensions.indexOf(extension) + EXTENSION_OFFSET
+
+  var callback = self._recieved[localId]
+  if (!callback) return
+  callback(localId, data.slice(1, data.length)) // remove extension number from front
+}
+
+Protocol.prototype.listAvailable = function () {
+  return this._available
+}
+
 Protocol.prototype._onhandshake = function (handshake) {
+  var self = this
   if (handshake.protocol !== 'hyperdrive' && handshake.protocol !== 'hypercore') {
     return this.destroy(new Error('Protocol must be hypercore'))
   }
+
+  if (handshake.extensions.length > 0) this._remoteExtensions = handshake.extensions
+
+  var common = handshake.extensions.filter(function (n) {
+    if (self._extensions.indexOf(n) === -1) return false
+    return true
+  })
+
+  common.forEach(function (ext) {
+    if (self._available.indexOf(ext) > -1) return
+    self._available.push(ext)
+  })
+
   this.remoteId = handshake.id
   this.emit('handshake')
 }
@@ -222,6 +294,7 @@ Protocol.prototype._parse = function (data) {
   }
 
   var type = data[0]
+  if (type >= EXTENSION_OFFSET) return this._route(type, data)
   if (type >= ENCODINGS.length) return // unknown message
 
   try {
--- a/lib/swarm.js
+++ b/lib/swarm.js
@@ -2,6 +2,7 @@ var protocol = require('./protocol')
 var debug = require('debug')('hyperdrive-swarm')
 
 var MAX_INFLIGHT_PER_PEER = 20
+var EXTENSION_OFFSET = 64
 
 module.exports = Swarm
 
@@ -14,6 +15,16 @@ function Swarm (core, opts) {
   this.feeds = []
   this.joined = {}
   this.kicking = false
+  this.extensions = []
+}
+
+Swarm.prototype._use = function (name) {
+  var index = this.extensions.indexOf(name)
+  if (index > -1) return EXTENSION_OFFSET + index
+  this.extensions.push(name)
+  // generate id using @mafintosh's proposed system https://github.com/mafintosh/hypercore/issues/4#issuecomment-170938932
+  var id = EXTENSION_OFFSET + (this.extensions.length - 1)
+  return id
 }
 
 Swarm.prototype._kick = function () {
@@ -139,7 +150,7 @@ Swarm.prototype.join = function (link) {
 
 Swarm.prototype.createStream = function () {
   var self = this
-  var peer = protocol()
+  var peer = protocol({ extensions: self.extensions })
 
   debug('new peer stream')
 
--- a/test/extensions.js
+++ b/test/extensions.js
@@ -1,0 +1,147 @@
+var tape = require('tape')
+var memdb = require('memdb')
+var hypercore = require('../')
+
+tape('use extension', function (t) {
+  var core = create()
+  var remote = create()
+
+  var ping = core.use('ping')
+  var remotePing = remote.use('ping')
+
+  var coreStream = core.createPeerStream()
+  var remoteStream = remote.createPeerStream()
+
+  remoteStream.pipe(coreStream).pipe(remoteStream)
+
+  coreStream.on('handshake', function () {
+    if (!coreStream.supports(ping)) return
+    coreStream.send(ping, new Buffer([0])) // send ping
+
+    remoteStream.receive(remotePing, function (extension, buf) {
+      remoteStream.send(remotePing, new Buffer([1]))
+    })
+
+    coreStream.receive(ping, function (extension, buf) {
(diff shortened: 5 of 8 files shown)


mafintosh:
@piedshag you wanna do a first pass at implementing the scheme we talked about in the issue?

piedshag:
absolutely i am working on it rn

piedshag:
done a few additions. definitely not ready. but my sandwich is.

piedshag:
which direction do you want to head in now? It finds common extensions on handshake and then calls `activate()` on that extension.

paulkernfeld:
@piedshag nice! I'll see if I can contribute anything useful to this today. I'll write down some thoughts as soon as I have them.

paulkernfeld:
I think this is definitely the right direction to go, and I like the idea of specifying the behavior with unit tests. I'm going to write a couple failing unit tests to demonstrate how I think the extension API should work. That should make it easier to agree on implementation. A couple miscellaneous thoughts:
- I think the extensions need to be stored in a list, so that the iteration order is well-defined. But maybe there's a JS trick here that I don't know about.
- I like your assertions in `Swarm.prototype._use`.

piedshag:
Hey @paulkernfeld. Most of the functionality is now implemented but currently there are no api's exposed to the extension. what do you think we should expose? 
eg.

``` js
ch.on('request', function (block) {
  var exts = getextensions()
  exts.forEach(function (ext) {
    ext.onRequest(block)
  })
})
```
^ just an example probably has no practical use

paulkernfeld:
@piedshag awesome :D

Here are my suggestions for an API to expose: piedshag/hypercore#1, please suggest improvements!

piedshag:
okey dokey

paulkernfeld:
Moving @piedshag's comment from piedshag/hypercore#1:
> Ok i see what you are doing with the ping extension. Looks good! So the api we expose should be similar to that of bittorrent protocol eg.

> onHandshake
> onExtendedHandshake
> and maybe some other event which is unique to hypercore

Relative to bittorrent-protocol:
- I renamed `onExtendedHandshake` to `extended` -- I worried that the reuse of the word "handshake" is a little confusing.
- I made a new event, `not_extended`, for when we find out that the remote does not support an extension that we support.

I actually can't think of any new hypercore-specific events that we would need. I am envisioning that the extensions API is only aware of hypercore's transport layer, not of hypercore feed logic (analogous to bittorrent-protocol).

paulkernfeld:
I'm going to write some more tests, e.g. to make sure that multiple extensions work.

piedshag:
sweet i am going to try and pass some tests

piedshag:
@paulkernfeld now passing all the tests you set. implemented the following
``` js
core.on('extended', function (ext) {
  // ext = extension shared by both pairs
})
core.on('not_extended', function (ext) {
  // ext = names of extensions
})
```

paulkernfeld:
Ah, thanks, I see you fixed my `'pong'` vs. `pong` bug!

> I don't think the extension callback is very useful. what do you think?
You mean the `extended` event, right? I like that event because you can use it to give the user a new instance of the extension when the protocol connects with another protocol. Is that a bad idea?

paulkernfeld:
@piedshag is there an effective way for me to help currently? I wouldn't mind writing additional tests, if that's useful.

piedshag:
The tests are great, but i think i need some guidance as to how the extention protocol will work. My aim was to create an extension which would allow for dynamic feeds to be created. I don't see how that is possible with the current api. Or am i over complicating it? should the dynamic feeds be directly implemented into the core and the extension protocol used for simpler things?

paulkernfeld:
Good point! You are right; the API that I'm exposing is lacking good hooks for doing a dynamic feed. My API has a nice *per-connection* interface (you could call it a channel), but it doesn't have an interface for doing stuff that relates to multiple peers. So we definitely need to add that.

As far as making a dynamic feed, I'm envisioning it in two parts. There are definitely other ways to do it, and I'd be interested to hear yours.
1. Implement BEP 44-like mutable values using this extension API
2. Add a separate module that uses these mutable values to store normal hypercore links. Then, to make the dynamic hypercore, look at the links and download the hypercore. When a link is updated, you can download the new hypercore parts that it points to. This separate module wouldn't use the extension API.

piedshag:
I think dynamic feeds and the extension api should be kept separate. I have a different approach to dynamic feeds similar to mafintosh's. I have a working demo [here](https://github.com/piedshag/webchain). It is v basic because i haven't put much effort into it. Basically the id of the stream is a ecc public key. All data broadcast into a swarm is verified based on the public key.

lib/protocol.js
```diff
@@ -18,7 +18,8 @@ var ENCODINGS = [
   messages.Unpause,
   messages.Request,
   messages.Response,
-  messages.Cancel
+  messages.Cancel,
+  messages.Extension
```
mafintosh: we shouldn't add a new message for extensions. if we use the approach described in https://github.com/mafintosh/hypercore/issues/4 all messages tagged >64 will be considered extension messages

mafintosh:
i was thinking about this last night and i think we should aim for a more low level api to expose the extension messages. based on the scheme i was discussing in https://github.com/mafintosh/hypercore/issues/4 extension messages are sent/received in just the same way as normal messages - just with a negotiated tag number that both ends do not have to agree on.

I therefore think we should have a simple stateless api that reflects that

``` js
var id = hypercore.use('my-extension') // returns some opaque integer id

var bool = stream.remoteSupports(id) // does the remote end support this extension?
stream.send(id, <some buffer>) // up to the user to encode their messages into buffers
stream.receive(id, function (buffer) {
  // up to the user to decode their messages
})
```

To create @piedshag's ping extension

```
var id = hypercore.use('ping')

if (stream.remoteSupports(id)) {
  stream.send(new Buffer([0])) // send ping
  stream.receive(id, function (buf) {
    if (buf[0] === 0) stream.send(id, new Buffer([1])) // send pong
    if (buf[0] === 1) console.log('got pong!')
  })
}
```

cc @feross - curious to hear his founds on o/

feross:
@mafintosh Can a user only get a `stream` once the remote handshake is received? If they can get it sooner, then `stream.remoteSupports(id)` won't work because that's sync.

piedshag:
Sounds great @mafintosh. Just a couple of questions. How are the unique extension ids generated? and when you say "messages are sent/received in just the same way as normal messages" does that mean the messages are not encoded?

mafintosh:
@feross they'll only be exposed after a handshake
@piedshag they are encoded however you want since it just sends buffers back and forth. If you to send structured messages you can just use json/protocol-buffers/something-else in your extension code

piedshag:
@mafintosh so how do we catch that message and send it to the appropriate extension? What do you think of my latest commit

piedshag:
@mafintosh i think the best way to do it is create a new message type for extensions

piedshag:
ping extension using current api
``` js
var ping = hypercore.use(ping)

if (stream.remoteSupports(ping)) {
  stream.send(ping, new Buffer([0])) // send ping
  stream.on(ping, function (buf) {
    if (buf[0] === 0) stream.send(ping, new Buffer([1])) // send pong
    if (buf[0] === 1) console.log('got pong!')
  })
}
```

index.js
```diff
@@ -62,6 +62,11 @@ Hypercore.prototype.add = function (opts) {
   return feed(this, null, opts)
 }
 
+Hypercore.prototype.use = function (ext) {
+  // WIP
+  return this.swarm._use(ext)
```
paulkernfeld: Is this supposed to take a string or the extension itself?

README.md
```diff
@@ -129,6 +129,10 @@ When a feed is being used hypercore will emit `interested` with the correspondin
 
 This is emitted when a feed is no longer being used.
 
+#### `core.use(extension)`
+
+Use an extension to the protocol. See section below.
```
paulkernfeld: I think it would be helpful to add an overview of the API, if we think that it is stable.

index.js
piedshag: mafintosh proposed a much lower level api see [here](https://github.com/mafintosh/hypercore/pull/5#issuecomment-171283614). Yeah it does accept a string now. I chose to go with mafintosh's api because i thought it was more practical

paulkernfeld: Ah, yeah, I get it. Cool!

lib/protocol.js
```diff
@@ -199,10 +204,51 @@ function Protocol (opts) {
 
 util.inherits(Protocol, duplexify)
 
+// extension api
+
+Protocol.prototype.remoteSupports = function (id) {
+  var self = this
+  if (!this._handshook) return
+
+  var offset = id - 64
```
paulkernfeld: It might be good to pull this 64 out as a constant so that it doesn't look like a [magic number](http://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad).

lib/protocol.js
```diff
@@ -222,6 +268,7 @@ Protocol.prototype._parse = function (data) {
   }
 
   var type = data[0]
+  if (type >= 64) return this.emit(type, data.slice(1, data.length))
```
paulkernfeld: I would use a constant for this 64 for readability.

lib/protocol.js
```diff
@@ -199,10 +204,51 @@ function Protocol (opts) {
...
+  return false
+}
+
+Protocol.prototype.listAvailable = function () {
+  return this._available
+}
+
+Protocol.prototype.send = function (id, buf) {
+  var self = this
+  if (!self.remoteSupports(id)) return
```
paulkernfeld: I would say that this should probably indicate if we're trying to send a message that the remote doesn't support. Otherwise, it's going to fail silently, which might be confusing.

What if this returns true or false depending on whether the remote supported the extension?

lib/protocol.js
piedshag: good idea. i will do it now

lib/protocol.js
```diff
@@ -199,10 +204,51 @@ function Protocol (opts) {
...
+
+  if (!Buffer.isBuffer(buf)) buf = new Buffer(buf)
+
+  var msg = new Buffer(buf.length + 1)
+  msg[0] = id
```
paulkernfeld: I don't think that we should currently be worried about users having more than 192 extensions, but should we put a thing in the code to check that that doesn't happen? E.g. assert that id < 255?

lib/swarm.js
```diff
@@ -14,6 +14,15 @@ function Swarm (core, opts) {
...
   this.joined = {}
   this.kicking = false
+  this.extensions = []
+}
+
+Swarm.prototype._use = function (name) {
+  if (this.extensions.indexOf(name) > -1) throw new Error('extension already exists')
+  this.extensions.push(name)
+  // generate id using @mafintosh's proposed system https://github.com/mafintosh/hypercore/issues/4#issuecomment-170938932
+  var id = 64 + this.extensions.indexOf(name)
```
paulkernfeld: This is another place where I think a constant would be nice.

lib/protocol.js
piedshag: @paulkernfeld another good idea. will do now

lib/protocol.js
piedshag: ok will do

lib/swarm.js
piedshag: done

paulkernfeld:
This looks great! It is a lot better than the API that I suggested, and the implementation seems simple and well-tested. The only thing that I'm not seeing is, don't we need to store the extension IDs that the remote is sending?

test/extensions.js
```diff
@@ -0,0 +1,133 @@
...
+
+  var coreStream = core.createPeerStream()
+  var remoteStream = remote.createPeerStream()
+
+  remoteStream.pipe(coreStream).pipe(remoteStream)
+
+  coreStream.on('handshake', function () {
+    coreStream.send(ping, new Buffer([0]))
+
+    remoteStream.on(ping, function (buf) {
```
paulkernfeld: I think that this test is "cheating" a little bit because we're using the `ping` variable to listen on the remote stream, but we got the value of `ping` from the core stream. IRL, we wouldn't be able to do this since they would be running on different machines.

piedshag:
@paulkernfeld yeah definitely. I haven't tested when the extensions are placed in a different order hence different id's. I will work on this now and get back to you. I have implemented your improvements you will see in teh next commit.

test/extensions.js
piedshag: lol yeah @paulkernfeld i have fixed it in latest commit. what do you think of it?

paulkernfeld:
This commit looks good to me!

lib/protocol.js
mafintosh: @paulkernfeld the id is technically a varint allowing us to support all the extensions in the world. I was just lazy when i first wrote thid

lib/swarm.js
```diff
@@ -14,6 +14,16 @@ function Swarm (core, opts) {
   this.feeds = []
   this.joined = {}
   this.kicking = false
+  this.extensions = []
+}
+
+Swarm.prototype._use = function (name) {
+  var index = this.extensions.indexOf(name)
+  if (index > -1) throw new Error('extension already exists')
```
mafintosh: Return 64 + i instead of throwing here

piedshag:
@mafintosh is there anything else i need to do for this to be merged?

mafintosh:
@piedshag i'll give this a final review tomorrow and get back to you.

mafintosh:
@piedshag in general it looked good the previous time i looked through. i'll give extra attention to this though as it it's gonna be exposed on the network.

btw thanks to both you and @paulkernfeld for taking the time to help with this

piedshag:
@mafintosh no problem, i enjoyed working on it. Just a quick question. Is hypercore going to be used as a substitute for hyperdrive or will hypercore be integrated into hyperdrive?

mafintosh:
@piedshag hypercore will used in hyperdrive. hyperdrive will act as a file sharing layer on top of this module.

mafintosh:
@piedshag i split it out into a separate thing so it would be easier to use for non-file related data you wanna distribute

piedshag:
@mafintosh ahhh ok. I noticed that hyperdrive duplicates most of the logic contained in hypercore. It will be good when we integrate hypercore in

mafintosh:
@piedshag yep. hopefully hyperdrive will only be a few 100 lines when we're done :)

package.json
```diff
@@ -24,6 +24,7 @@
   },
   "devDependencies": {
     "memdb": "^1.3.1",
+    "sodium-signatures": "^1.1.0",
```
mafintosh: could you make the tests work without sodium? its a pretty big dep

test/extensions.js
```diff
@@ -0,0 +1,184 @@
...
+
+    remoteStream.receive(remotePing, function (extension, buf) {
+      t.same(buf, new Buffer([0]), 'multiple extension message routing works')
+      t.end()
+    })
+  })
+})
+
+tape('sign and verify example', function (t) {
+  var keys = signatures.keyPair()
```
mafintosh: see above

mafintosh:
@piedshag looking good! see my final minor comment. also could rebase this into a single commit?

piedshag:
@mafintosh. lol i had about 3 attempts at rebasing it. Hope this is sufficient.

mafintosh:
Landed in 58a9be3

mafintosh:
@piedshag github said i closed this but its actually merged. i just squashed the two commits into a single one for you and changed the commit message.

PR #6, opened 2016-01-23
Extension api v2

piedshag:
As we discussed in [irc](https://botbot.me/freenode/dat/2016-01-22/?msg=58513579&page=2) the current api is very low level. This pr makes the api available on a higher level very similar to feross' [bittorrent-protocol](https://github.com/feross/bittorrent-protocol/). I have created an example extension [here](https://github.com/piedshag/he-chat). What are your thoughts on this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "uint64be": "^1.0.1"
   },
   "devDependencies": {
+    "he-chat": "^1.0.0",
     "memdb": "^1.3.1",
     "standard": "^5.4.1",
     "tape": "^4.2.2"
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules
+.DS_Store
--- a/README.md
+++ b/README.md
@@ -129,6 +129,10 @@ When a feed is being used hypercore will emit `interested` with the correspondin
 
 This is emitted when a feed is no longer being used.
 
+#### `core.use(extension, cb)`
+
+When we receive extended handshake cb will be called with (id, peer). See [he-chat](https://github.com/piedshag/he-chat)
+
 ## Feeds
 
 Everytime you write a stream of data to hypercore it gets added to an underlying binary feed. Feeds give you more low-level access to the data stored through the following api.
--- a/index.js
+++ b/index.js
@@ -62,6 +62,11 @@ Hypercore.prototype.add = function (opts) {
   return feed(this, null, opts)
 }
 
+Hypercore.prototype.use = function (ext, cb) {
+  // WIP
+  return this.swarm._use(ext, cb)
+}
+
 Hypercore.prototype._close = function (link) {
   var id = link.toString('hex')
   delete this._opened[id]
--- a/index.js
+++ b/index.js
@@ -62,9 +62,9 @@ Hypercore.prototype.add = function (opts) {
   return feed(this, null, opts)
 }
 
-Hypercore.prototype.use = function (ext, cb) {
+Hypercore.prototype.use = function (extension) {
   // WIP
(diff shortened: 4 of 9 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
     "uint64be": "^1.0.1"
   },
   "devDependencies": {
+    "he-chat": "^1.0.0",
```
paulkernfeld: This dependency is in a weird direction from my perspective. I would think that the extensions would depend on hypercore, rather than the other way around.

paulkernfeld:
I like this higher-level extension idea and API; I think this could make it easier for people to extend hypercore. I don't have a strong opinion on whether this is the right way to implement this. For example, I could imagine that the higher-level extension API might be implemented in a separate package that depends on hypercore. I think that decision really depends on the tradeoff between maintaining simplicity in hypercore vs. providing a convenient API to extension authors.

Also, it looks like this needs to be rebased.

mafintosh:
I've tweaked the low-level api a bit. See the docs here, https://github.com/mafintosh/hypercore#extension-api - I agree with @paulkernfeld's comment. Gonna think about this for a bit.

mafintosh:
You could write he-chat like this atm actually using the current extension api

```
function hechat (core) {
  core.use('he-chat')
  // .. do stuff
  return {
    broadcast: function (message) {
      for (var i = 0; i < peers.length; i++) {
        peers[i].send('he-chat', message)
      }
    } 
  }
}
```

That might be even more flexible

piedshag:
> I think that decision really depends on the tradeoff between maintaining simplicity in hypercore vs. providing a convenient API to extension authors.

@paulkernfeld yeah i agree. I think that being able to develop extensions in external modules would make the api more convenient though. @mafintosh i like your enhancements to the pr it makes it much simpler. I had no idea you could do that with the current codebase, looks great to me as long as you can develop it in an external module. How does the broadcast function get access to the peers?

mafintosh:
@piedshag good point. what if i added a `core.peers` array with all the current peers?

piedshag:
@mafintosh yeah that would be great. It would make the api a lot more functional at a higher level. I think you can already access `core.swarm.peers` but `core.peers` would be more convenient

mafintosh:
okay so here is my proposal. we add `core.peers` and `on('peer', peer) and encourage the following pattern for extensions

```
function broadcastExtension (core) {
  var ext = new events.EventEmitter()

  core.use('broadcast')
  core.on('peer', function (peer) { // will be emitted when a peer stream is created
    peer.on('broadcast', function (message) {
      ext.emit('broadcast', message)
    })
  })

  ext.broadcast = function (m) {
    core.peers.forEach(function (peer) {
      peer.send('broadcast', m)
    })
  }

  return ext 
}
```

thoughts?

piedshag:
@mafintosh yup sounds good. I think the extension should be encouraged to check that the peer supports the extension before broadcasting, like so.
``` js
function broadcastExtension (core) {
  var ext = new events.EventEmitter()

  core.use('broadcast')
  core.on('peer', function (peer) { // will be emitted when a peer stream is handshaked
    peer.on('broadcast', function (message) {
      ext.emit('broadcast', message)
    })
  })

  ext.broadcast = function (m) {
    core.peers.forEach(function (peer) {
      if (!peer.remoteSupports('broadcast')) return
      peer.send('broadcast', m)
    })
  }

  return ext 
}
```

mafintosh:
@piedshag yep. the .send api discards unsupported messages though

piedshag:
@mafintosh ohh ok lol. not to worry then

mafintosh:
closing this since we agree on the o/ apprach

PR #12, opened 2016-03-29
added hash and sign methods to feeds

lukeburns:
feed.hash() returns the hash of the merkle roots or null if there are no roots
feed.sign(message) returns the signed message or null if there is no secret key


mafintosh:
i'm closing this because we've made a bunch of changes since this was opened. if this is still relevant please re-open and i'll take a look :)

PR #13, opened 2016-04-16
fixes for read and write streams, electron fix

substack:
This patch fixes create{Read,Write}Stream and handles an issue under electron.


mafintosh:
thanks. i'm gonna add tests for the stream interface as well.

mafintosh:
2.0.1

PR #22, opened 2016-06-01
fix default key encoding

juliangruber:
The db i passed to hyperdrive/hypercore had `{keyEncoding: bytewise}`, so this was causing bugs.

The comment I removed said "binary" encoding, but I the keys written were all strings so I used `utf8` for simplicity.

### DIFF
--- a/test/helpers/create.js
+++ b/test/helpers/create.js
@@ -2,5 +2,6 @@ var hypercore = require('../../')
 var memdb = require('memdb')
 
 module.exports = function create () {
+  // ensure hypercore overwrites the db's default encoding
   return hypercore(memdb({keyEncoding: 'json'}))
 }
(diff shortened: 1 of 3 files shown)


test/helpers/create.js
```diff
@@ -2,5 +2,5 @@ var hypercore = require('../../')
 var memdb = require('memdb')
 
 module.exports = function create () {
-  return hypercore(memdb())
+  return hypercore(memdb({keyEncoding: 'json'}))
```
mafintosh: i'm curious why this is needed?

juliangruber: to make the tests prove that you can supply any key encoding. should i add a comment?

mafintosh: ah cool! yea that'd be great.

juliangruber: done!

PR #28, opened 2016-06-10
swap out crypto with specific modules

substack:
This patch reduces the browser payload.

Before:

```
$ browserify .|wc -c
1192909
```

After:

```
$ browserify .|wc -c
790215
```


substack:
should help with https://github.com/mafintosh/hypercore/issues/26

mafintosh:
4.2.6

PR #30, opened 2016-07-01
get block from hash method

substack:
This patch adds `getBlockFromHash()` to look up the block index from a rolling head hash. This is needed upstream by hyperdrive to implement checkout.


clkao:
Can you add some tests for the new api?

clkao:
Ok now travis runs with latest node v6 and this should be green. @substack can you rebase to trigger the tests?

substack:
This is the wrong approach for this feature.

PR #35, opened 2016-07-17
Fix crash when trying to close storage w/o a .close method

paulkernfeld:


mafintosh:
thanks! will release when i get to my laptop

PR #36, opened 2016-07-21
add options.values to list() method

pfraze:
This PR adds the `hypercore.list({values: true})` usecase, which enables an API consumer to quickly read listing metadata.

In my case, I needed to list all the archives that I own. This PR lets me quickly fetch all archives along with their secret keys. If the secret key is populated, I know I'm the owner.


pfraze:
Is node 0.10 really supported by hypercore? I use Buffer.compare in the tests, which was added in 0.11

mafintosh:
out in 4.8.0

PR #49, opened 2016-11-23
Add unreplicate()

pfrazee:
Closes https://github.com/mafintosh/hypercore/issues/48

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,12 @@ Options include:
 }
 ```
 
+#### `core.unreplicate(feed, [stream])`
+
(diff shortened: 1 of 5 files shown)


pfrazee:
The method signature here differs from hyperdrive's, so, if this is right, we'll need to update hyperdrive's usage.

pfrazee:
weird, I wonder why standard didnt run on my tests. Investigating...

pfrazee:
Whoops, my es6 is leaking

pfrazee:
Ok ready for review

mafintosh:
4.15.0

PR #50, opened 2016-12-14
wip: non live replication

mafintosh:
per default replication streams will exit when the current snapshot has been synced between both peers (hyperlog style).

``` js
var hypercore = require('./')
var memdb = require('memdb')

var core1 = hypercore(memdb())
var core2 = hypercore(memdb())

var feed1 = core1.createFeed()
var feed2 = core2.createFeed(feed1.key)

feed1.append([
  'how',
  'are',
  'you',
  'doing?',
  'i',
  'am',
  'fine',
  'thanks',
  'what',
  'about',
  'you?'
])

feed1.flush(function () {
  var s1 = feed1.replicate()
  var s2 = feed2.replicate()

  s1.pipe(s2).pipe(s1)

  s1.on('end', function () {
    feed2.get(0, function (err, buf) {
      console.log('->', buf.toString())
    })
  })
})
```

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -45,6 +45,7 @@ function Peer (feed) {
   this._tick = 0
   this._index = -1
   this._blocks = 0
+  this._once = true
 
   this.uploading = true
   this.downloading = true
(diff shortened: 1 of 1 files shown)


mafintosh:
fixed in master

PR #57, opened 2017-01-16
Add verify-on-read (WIP)

pfrazee:
Will close https://github.com/mafintosh/hypercore/issues/55

Still TODO:

 - [x] Tests
 - [x] How should the error condition be handled during replication? Is there a way to handle a failed read-verify other than destroying the connection? Perhaps send an unhave message?
 - [x] Update the local bitfield on failed verificaiton

### DIFF
--- a/lib/feed.js
+++ b/lib/feed.js
@@ -195,7 +195,7 @@ Feed.prototype.get = function (block, opts, cb) { // TODO: on static feeds retur
         if (err) return cb(err)
         var blockHash = hash.data(value)
         if (!equals(blockHash, node.hash)) {
-          // TODO update local bitfield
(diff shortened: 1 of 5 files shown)


lib/feed.js
```diff
@@ -170,16 +170,43 @@ Feed.prototype.has = function (block) {
...
+          self.bitfield.set(block, false) // remove from the local bitfield
+          self._sync()
+          cb(new Error('Block not found'))
+        } else {
+          done(null, value)
+        }
+      })
+    } else {
+      done(null, value)
+    }
```
mafintosh: you can same a lot of indentation by doing `if (!opts.verify) return done(null, value)` first

pfrazee: fixed

PR #63, opened 2017-01-17
createFeed use key over opts.key

joehand:
Fixes https://github.com/mafintosh/hypercore/issues/62

* Takes `key` over `opts.key`
* If there is not a `key`, uses `opts.key` (and converts to buffer if its a string)
* Tests


mafintosh:
4.19.2

PR #64, opened 2017-01-17
Add {wait} opt to feed.get()

pfrazee:
This PR makes it possible to do `feed.get(block, {noWait: true}, ...)`. If the block isnt available, rather than queue it for download, it will callback immediately with a notFound error.

This is needed for https://github.com/joehand/hypercloud/issues/50

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ Retrieve a block of data from the feed. If the block is not locally-available, t
 ``` js
 {
   verify: false // verify the data against the feed checksum, and fail the get() if !==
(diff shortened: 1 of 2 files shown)


mafintosh:
I'd prefer you called this `{wait: false}` instead. Other than that 👍

pfrazee:
Done

mafintosh:
4.19.0

pfrazee:
Thanks!

juliangruber:
nice!

PR #65, opened 2017-01-18
Fix: Pass get options through _openAndGet

pfrazee:
This fixes an edgecase where, if the feed was not open yet, the opts would not get correctly passed to the get() after open


mafintosh:
4.19.1

pfrazee:
thanks!

PR #86, opened 2017-04-12
add 'complete' event for download finished

karissa:
What if we add back the download-finished event but call it something else, like 'complete'? Any better ideas?


mafintosh:
5.11.0 has `sync` which is basically what you add here but using an internal index so it's more efficient (i.e. no need to look at all the data)

karissa:
Great!!

Issue #87, opened 2017-04-12
'want' of null bug when having multiple feeds 


karissa:
Was able to produce a reproducible test case by running `npm test` on this branch. Seems to happen when you have two hypercores in the same swarm. https://github.com/mafintosh/hypercore-archiver/pull/27/files#diff-bdf3fde96319b43abcc38201204c47ffR97

```
/Users/karissa/node_modules/hypercore-archiver/node_modules/hypercore/lib/replicate.js:269
  this.stream.want({start: 0}) // TODO: don't just subscribe to *EVERYTHING* hehe
             ^

TypeError: Cannot read property 'want' of null
    at Peer.ready (/Users/karissa/node_modules/hypercore-archiver/node_modules/hypercore/lib/replicate.js:269:14)
    at /Users/karissa/node_modules/hypercore-archiver/node_modules/hypercore/lib/replicate.js:24:10
    at /Users/karissa/node_modules/hypercore-archiver/node_modules/hypercore/index.js:112:15
    at apply (/Users/karissa/node_modules/hypercore-archiver/node_modules/thunky/index.js:44:12)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

mafintosh:
Fixed by you in #92

PR #92, opened 2017-04-17
if no stream on ready and have, don't try to do anything

karissa:
This is fixing #87 now, but I wonder if there's something else deeper to do here that is slightly over my head..


mafintosh:
This is actually pretty spot on. The `have` one can safely be ignored (new blocks getting added before the stream is ready). The `ready` one is because the stream has been destroyed. I've commented on your archiver PR as to why that is happening.

mafintosh:
Out in 5.11.1

PR #97, opened 2017-05-08
Add feed.cancel(index)

mafintosh:


index.js
```diff
@@ -341,7 +342,40 @@ Feed.prototype.put = function (index, data, proof, cb) {
...
+      this.undownload(sel)
+    }
+  }
+
+  // defer the last part until after ready as .get does that as well
+  if (this.opened) this._cancel(start, end)
+  else this._readyAndCancel(start, end)
+}
+
+Feed.prototype._cancel = function (start, end) {
```
juliangruber: maybe rename to `._cancelAfterOpen` so you can get rid of the comment

index.js
```diff
@@ -341,7 +342,40 @@ Feed.prototype.put = function (index, data, proof, cb) {
...
+  // expect this to be run after feed has fully opened
+  var i = 0
+
+  for (i = start; i < end; i++) {
+    this._reserved.set(i, false) // TODO: send cancel message if set (requires it to be cleared on put)
+  }
+
+  for (i = this._waiting.length - 1; i >= 0; i--) {
+    var w = this._waiting[i]
+    if ((start <= w.start && w.end < end) || (start <= w.index && w.index < end)) {
```
juliangruber: why `< end` here but `<= end` above?

index.js
mafintosh: good point

mafintosh: actually, we expect all the _get/_put etc methods to run after open so that is implied. let me just remove the comment

index.js
mafintosh: bug! nice catch. gonna add a test for this case

mafintosh:
Gonna add docs once I test this a bit in practice

PR #101, opened 2017-06-05
downloaded function

e-e-e:
This is perhaps more complicated than I initially thought, but it is much more efficient than iterating over the individual bits. Basically I get the total by iterating over the byte range and tally the number of ones using a look up table. I tried to mirror your style as much as possible, but let me know if you have any suggestions how to make it better.

As for adding range option to `has` what were your thoughts. A boolean response if all bits are downloaded? If so we can add that easily with the help of this function too. I am just unsure how you were thinking to handle the args, would it expect an object {start: 0, end: n} or just assume that if you include a second arg that its a range, other wise just check individually. Let me know, and I will add it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ node_modules
 sandbox
 bench/cores
 tmp
+
+*.log
--- a/README.md
+++ b/README.md
@@ -154,6 +154,12 @@ feed will be marked for download for you when the feed is created.
 
 Cancel a previous download request.
 
+#### `var number = feed.downloaded([start], [end])`
+
+Returns total number of downloaded blocks within range.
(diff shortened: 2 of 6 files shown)


mafintosh:
This is great! You can actually make this function even faster by using the bitfield index as well to skip over large sections of `1`/`0` bits but this is fine for now.

mafintosh:
I think you should make another PR for `.has(start, [end])` btw. Then we can land this faster

mafintosh:
Can I get you to add a test that uses the feed interface as well and tests the defaults?

e-e-e:
I would be super curious how your bitfield index works - i did not quite get my head around it all.

mafintosh:
6.4.0

PR #103, opened 2017-06-10
fix issue with bitfield.total() counting full bytes as -1

e-e-e:
There as a silly error where 255 was counted as -1. I added a test for large byte ranges.
This should cover it now.


mafintosh:
On my phone. Will release when at laptop

mafintosh:
Now released. 6.5.1

PR #104, opened 2017-06-12
Add download opt to replicate

pfrazee:
This makes it possible to enter the swarm while only uploading data, using the `download: false` on `replicate()`.

Adding an `upload` opt to replicate will take more work and I don't need that feature right now, so I'm just submitting this.

### DIFF
--- a/test/replicate.js
+++ b/test/replicate.js
@@ -551,7 +551,7 @@ tape('replicate no download', function (t) {
   feed.append(['a', 'b', 'c', 'd', 'e'], function () {
     var clone = create(feed.key)
 
-    clone.get(0, function (err) {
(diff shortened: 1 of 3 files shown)


test/replicate.js
```diff
@@ -545,6 +545,26 @@ tape('feed has a large range', function (t) {
...
+    var clone = create(feed.key)
+
+    clone.get(0, function () {
+      t.fail('Data was received')
+    })
+
+    var stream = feed.replicate({live: true})
+    stream.pipe(clone.replicate({live: true, download: false})).pipe(stream)
+
+    setTimeout(() => {
```
mafintosh: you know what i'm gonna comment on here, don't you?

PR #106, opened 2017-07-13
hash -> crypto

lukeburns:
removed sodium-signatures dep


mafintosh:
Beautiful!

mafintosh:
Best thing about this is that it is a patch update and we are now only one patch bump away from 6.6.6

lukeburns:
Hehe gotta make the next patch a good one

Issue #108, opened 2017-07-19
do not re-write keys and stuff on open


mafintosh:
if a key is read from storage do not rewrite it here, https://github.com/mafintosh/hypercore/blob/master/index.js#L219 (same goes for secret key)

PR #110, opened 2017-07-26
Don't rewrite keypair on open when aren't new

noffle:
My first PR! Addresses #108.

### DIFF
--- a/index.js
+++ b/index.js
@@ -213,8 +213,8 @@ Feed.prototype._open = function (cb) {
         self._storeSecretKey = false
       }
 
-      var shouldWriteKey = self.key && generatedKey
-      var shouldWriteSecretKey = self._storeSecretKey && generatedKey
+      var shouldWriteKey = generatedKey || !safeBufferEquals(self.key, state.key)
+      var shouldWriteSecretKey = (self._storeSecretKey && generatedKey) || !safeBufferEquals(self.secretKey, state.secretKey)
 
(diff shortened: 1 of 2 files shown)


pfrazee:
huzzah!

noffle:
Tweaks: `standard` doesn't like my human-readable formatting on the `missing` list.

mafintosh:
Would be cool to add a test for that but I can do that for you :)

noffle:
If the storage key isn't the same as the generated or constructor key, [it looks like an error case](https://github.com/mafintosh/hypercore/blob/master/index.js#L184). I'm not sure I understand the case where the constructor offers a secret key but a fresh public key is generated -- aren't keypairs codependent?

mafintosh:
@noffle i meant the case where you are creating a fresh hypercore but providing both key+secretKey in the constructor. Then generated will be false but it needs to write them to storage still.

mafintosh:
@noffle could be that your code already works for that. thats why we need a test for that case :)

mafintosh:
here is an example of an existing storage test, https://github.com/mafintosh/hypercore/blob/master/test/basic.js#L70

noffle:
I think I understand now. Thanks for explaining @mafintosh. :heart:

noffle:
I think this fella is ready to :ship:!

mafintosh:
\o/ 6.6.7

Issue #111, opened 2017-08-03
add .head or .latest


mafintosh:
to get the last inserted item (feed.get(feed.length - 1, cb))

PR #112, opened 2017-08-04
Add .head function

noffle:
Addresses #111.

### DIFF
--- a/index.js
+++ b/index.js
@@ -789,8 +789,12 @@ Feed.prototype.has = function (start, end) {
 
 Feed.prototype.head = function (opts, cb) {
   if (typeof opts === 'function') return this.head({}, opts)
-  if (this.length === 0) cb(new Error('feed is empty'))
-  else this.get(this.length - 1, cb)
+  var self = this
+  this._ready(function (err) {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -781,6 +781,12 @@ Feed.prototype.has = function (start, end) {
   return total === this.bitfield.total(start, end)
 }
 
+Feed.prototype.head = function (opts, cb) {
+  if (typeof opts === 'function') return this.head({}, opts)
+  if (this.length === 0) cb(new Error('feed is empty'))
+  else this.get(this.length - 1, cb)
```
mafintosh: you should wrap this in .ready so .length is always loaded :) can we make it return null, null if there is no head?

mafintosh: nevermind the null, null - i guess we return an error in .get as not found as well

noffle: Do you think it would be confusing to have `.get` return immediately but `.head` to wait for `ready`?

mafintosh: .get waits for ready right now

noffle: Understood :bulb:

mafintosh:
6.7.0

PR #115, opened 2017-08-17
add defaults with support for indexing

joehand:
Part one of fixing our indexing issue. Adds `feed.defaults(opts)` to change the indexing option. 

Are there any options I should add in here? Didn't see anything that jumped out.


joehand:
Note to self, before merge: I should add docs for the indexing option with this PR too.

mafintosh:
Closing cause we have modules for this and stuff now :)

PR #126, opened 2017-12-20
un-lazily load replication requirements

garbados:
If the filesystem has been mocked (such as by [mock-fs](https://www.npmjs.com/package/mock-fs)), hypercore cannot lazily load its replication dependencies because they will appear to not exist.

This PR eliminates the lazy loading and requires the replication dependencies when the script is first loaded.


mafintosh:
is that because mock-fs expects all requires to happen in the first tick?

mafintosh:
fixed at some point \o/

PR #127, opened 2017-12-22
bit twiddling  ~10% perf boost 

d4tocchini:
From a casual look, bench results appear boosted ~10% (latest greatest macbook pro, node v8.9.1) 

@mafintosh, i can add back that inherits dep... can't help myseful :) the abundance of hidden crap these days in forgotten package chains has me axing all but the exceptional dep - this being one such rare gem, thanks!


noffle:
Ha, I didn't know about the `| 0` trick for integer flooring.

mafintosh:
@d4tocchini nice! My only concern is that bit fiddling only works for 32 bit numbers in js

d4tocchini:
@mafintosh, indeed, any numericals in there that shouldn't be assumed within 32bits? or less than `2^31 - 1` as that 32nd bit gets hairy in JS.  tried to keep it as an extension to what you had already established in the bitfield.  

As for big int, there are some interesting native module / wasm possibilities as of late that prob should be explored even if not for extra bits. And surprisingly, packing & unpacking pairs into 48bit uint values is actually quite fast in JS, with a few tricks it can be done without division!

d4tocchini:
Getting all ops as close as possible to constant time helps perf but also crytpographic integrity :)  There's something addictive about bit opping & monomorphizing all the things, not just for the huge perf wins... I tried to restrain myself, but benchmarking default JS idioms like everything in`Math.` & `Array.prototype.` against even naive implementations and it's apparent we've grown up on Monsato corn syrup & dont know the better.  We shouldn't be able to crush "native" method perf by often an order of magnitude.  Even math operators like `<` can be edged out in some cases, see rabbit-hole bench snippet below:

```
const int31 = 31 | 0;
const POW31 = 2147483648;
  
  // x > 0
  function gt_0_js(x)      {return x > 0}
  function gt_0_js_cast(x) {return ((x|0) > 0) | 0}
  function gt_0_a(v)       {return -v >>> 31}
  function gt_0_b(v)       {return (~v + 1) >>> 31}
  function gt_0_v1(v)      {return -v >>> int31}  
  function gt_0_v2(v)      {return -(v|0) >>> int31 }
  function gt_0_v3(v)      {return v ? (v >>> int31) ^ 1 : v}
  function gt_0_v4(v)      {return (v) ? ((v >>> int31) ? 0 : 1) : 0}
  function gt_0_v5(v)      {return (v >>> int31) ? 0 : (v ? 1 : 0)}
  function gt_0_v6(v)      {return (v) ? ((v & POW31) ? 0 : 1) : 0}
  function gt_0_v7(v)      {return (v & POW31) ? 0 : (v ? 1 : 0)}

  // x < 0
  function lt_0_js(x)      {return x < 0}
  function lt_0_js_cast(x) {return ((x|0) < 0) | 0}
  function lt_0(v)         {return v >>> 31}
  function lt_0_v1(v)      {return v >>> 31}
  function lt_0_v2(v)      {return (v|0) >>> int31}
```

@noffle, the `|0` in the benchmarks is there not to cast-as-int & floor, but to test it's impact on perf in general.  In some environments, it leads to substantial perf gains. i think partly b/c it was the mechanism in `asm` to strongly type as int and stuck around in the compilers

telamon:
I see that this is an  old issue but I stumbled upon it today by accident and would like to suggest some closure. 

Performance optimizations at the cost of readability and maintenance is not really a good trade-off.
On the other hand I see no reason on why there shoulnd't exist a source-code preprocessor that would apply this kind of optimizations through an optional precompilation step. (awesomesaucify.js)

As for high-performing hypercores I'm excitingly following the rust implementation attempt.

mafintosh:

mafintosh:

PR #128, opened 2018-01-07
Detect mismatching keypair

xloem:
I managed to create feed storage with signatures signed by multiple different keys!  The software didn't complain at all, but nobody could download from me.

This patch prevents that from beginning.

### DIFF
--- a/index.js
+++ b/index.js
@@ -217,6 +217,12 @@ Feed.prototype._open = function (cb) {
     if (state.key) self.key = state.key
     if (state.secretKey) self.secretKey = state.secretKey
 
+    if (self.key && self.secretKey) {
+      var challenge = new Buffer('');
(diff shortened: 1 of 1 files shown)


mafintosh:
@xloem standard is complaining, mind fixing?

index.js
```diff
@@ -217,6 +217,12 @@ Feed.prototype._open = function (cb) {
     if (state.key) self.key = state.key
     if (state.secretKey) self.secretKey = state.secretKey
 
+    if (self.key && self.secretKey) {
```
mafintosh: Adding a comment in front of this if saying that this is to test that the public key corresponds to the secret would be nice to help future collabs grok this :)

mafintosh:
👍 6.11.1

PR #129, opened 2018-01-08
Add signature() and verify() functions to API

### LINKED ISSUE #125: Provide Access to Hashes for Secure Trusted Timestamping
API functions providing a purported hash dependent on all preceding data at a single point in the stream, and for verifying such a hash given the availability of all this data, would allow for the data to be secured with a trusted timestamp.

This could allow to prove that data was not tampered with after the timestamp, and verify this proof, ala https://en.wikipedia.org/wiki/Trusted_timestamping .  But so far access to such a hash is not yet provided by the API.

xloem:
Allows user to retrieve a small signed proof of a data block and to verify such proofs.

Fixes #125

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,30 @@ feed will be marked for download for you when the feed is created.
 
 Cancel a previous download request.
 
+#### `feed.signature([index], callback)`
+
+Get a signature proving the correctness of the block at index, or the whole stream.
+
+Callback is called with `(err, signature)`.
+The signature has the following properties:
+``` js
+{
+  index: lastSignedBlock,
+  signature: Buffer
+}
+```
+
+#### `feed.verify(index, signature, callback)`
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -162,6 +162,30 @@ feed will be marked for download for you when the feed is created.
...
+
+#### `feed.verify(index, signature, callback)`
+
+Verify a signature is correct for the data up to index, which must be the last signed
+block associated with the signature.
+
+Callback is called with `(err, success)` where success is true only if the signature is
+correct.
+
+It is fine if the passed signature is unfamiliar; it is recalculated for verification,
```
mafintosh: what does `unfamiliar` mean here?

xloem: It means `not present in local storage`.

I might have had the details of the backend floating around in my head too strongly when I wrote this =)  Happy to pull it out.

mafintosh:
6.12.0 \o/

PR #130, opened 2018-01-08
Add support for acknowledgments

emilbayes:
I want to make a scheme where I can be reasonably certain that n-peers have received a block before I communicate any confidence that a block is safely "committed". This PR goes with the new flag in the protocol handshake and asks peers to acknowledge any blocks they receive. Of course this does not make any guarantees about byzantine peers, so you should only trust the ack if you trust the peer. 

Still needs more tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,7 @@ Options include:
 ``` js
 {
   live: false, // keep replicating after all remote data has been downloaded?
+  ack: false, // set to true to get explicit acknowledgement when a peer has written a block
   download: true, // download data from peers?
   encrypt: true // encrypt the data sent using the hypercore key pair
 }
--- a/index.js
+++ b/index.js
@@ -799,7 +799,7 @@ Feed.prototype._verifyRootsAndWrite = function (index, data, top, proof, nodes, 
 Feed.prototype._announce = function (message, from) {
   for (var i = 0; i < this.peers.length; i++) {
     var peer = this.peers[i]
-    if (peer !== from) peer.have(message)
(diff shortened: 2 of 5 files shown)


pfrazee:
I'm 👍  for the idea of this PR (haven't done code review).

(An aside on why I'm interested in this.) I'm personally still debating whether multiwriter is the right solution for handling multiple devices, or if we should only be using multiwriter for collaboration between multiple *people*. If we had a 'safe commit' protocol, we could have users reuse the same publishingkeys between devices, and then require checkins with managernodes before publishing updates to a log. This has both upsides and downsides.

Upsides:

 - Easier to communicate to users how to manage their keys; they can just copy the keyfiles from one device to another.
 - With multiwriter, there's a master hypercore which appoints new writers. If we use multiwriter to manage many users' devices, then we'll need a solution for what happens if the master-core is lost. (That's a downside for using multiwriter to manage user devices.)

Downsides:

 - You can make writes offline, but you can't publish/share the data until you've synced with your managernode. If you make writes that conflict, you'll need to roll them back and reapply them. (You can use a 'conflict' states like in multiwriter when writes conflict.)
 - More conceptual complexity, because there are now two syncing systems: safe commits and multiwriter.
 - Depends on managernodes, which hurts the automagic decentralization.

I'm not convinced of my argument but I wanted to record it.

martinheidegger:
@emilbayes what is missing here?

emilbayes:
@martinheidegger this looks pretty stale, but it is also lacking tests

mafintosh:
landed in latest

PR #132, opened 2018-01-19
Added test for signature failure

xloem:
Verifies signature checking code fails for bad data.  This was previously missing from tests.


mafintosh:
Awesome!

PR #143, opened 2018-03-18
Accept an optional discoveryKey

javisantos:
I think is a good idea to have an optional discoveryKey


javisantos:
What happens when 2 peers have the same discoveryKey for differents Keys?

mafintosh:
Standard doesn't seem to like your test for some reason. Other than that I'm okay with making it configurable.

martinheidegger:
@mafintosh seems like @javisantos pushed fixes.

mafintosh:
Closing cause old and lots have changed (feel free to reopen again latest, feature makes sense)

PR #145, opened 2018-03-23
Fix: properly remove peers on stream close/end

pfrazee:
Pretty sure this was causing a memory leak. `Peer.onclose()` was never getting called, and so `Peer.destroy()` was only called on certain error conditions. Graceful closes were not getting covered!

Also had to move the removal from the peers list within `destroy()` because the peer may not be totally setup prior to `destroy()` occurring.

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -351,20 +351,22 @@ Peer.prototype.end = function () {
 }
 
 Peer.prototype.destroy = function (err) {
-  // make sure the peer is no longer being tracked
-  set.remove(this.feed.peers, this)
-
   // deconstruct any state
-  if (this._index === -1) return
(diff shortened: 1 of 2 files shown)


mafintosh:
This should be triggered by https://github.com/mafintosh/hypercore-protocol/blob/master/index.js#L237 - do you have a failing test case by chance?

pfrazee:
https://github.com/mafintosh/hypercore-protocol/blob/master/feed.js#L95 close() removes the feed prior to the codepath that runs the peer onclose()

mafintosh:
Fixed by https://github.com/mafintosh/hypercore-protocol/commit/4566ce9e6073f5af99f7d6df1962ab33a1e446b5 - Added your test case though :+1:

PR #147, opened 2018-03-27
Make sure the stream is open before using it (close #146)

### LINKED ISSUE #146: Cannot read property 'data' of null in replicate.js
I got this report from somebody building from source (https://github.com/beakerbrowser/beaker/issues/919) so there's a good chance theyre using the latest hypercore. Here's the error:

```
Uncaught Exception:
TypeError: Cannot read property 'data' of null
    at Request.onvalue [as_callback] ($PATH_TO_BEAKER/app/node_modules/hypercore/lib/replicate.js:135:19)
    at Request.callback ($PATH_TO_BEAKER/app/node_modules/random-access-storage/index.js:150:8)
    at onread ($PATH_TO_BEAKER/app/node_modules/hyperdrive/node_modules/random-access-file/index.js:80:31)
    at FSReqWrap.wrapper [as oncomplete] (fs.js:658:17)
```

In https://github.com/mafintosh/hypercore/commit/e7b2b51d2836c6c1808432c78649da84c3b13016#diff-c169e1e72632697264bcd930cc444020 we set `.stream` to null on close, and I think it's possible for some async work to finish after that (eg https://github.com/mafintosh/hypercore/blob/e7b2b51d2836c6c1808432c78649da84c3b13016/lib/replicate.js#L151). I think the solution is to just return early in https://github.com/mafintosh/hypercore/blob/e7b2b51d2836c6c1808432c78649da84c3b13016/lib/replicate.js#L118 if `self.stream` is falsy.

pfrazee:
Added a bunch of guards against using `.stream` if it's null, which could happen if a call is made after the connection is closed

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -61,6 +61,7 @@ function Peer (feed, opts) {
 }
 
 Peer.prototype.onwant = function () {
+  if (!this.stream) return
   // TODO: reply to the actual want context
   this.remoteWant = true
   var rle = this.feed.bitfield.compress()
@@ -117,6 +118,7 @@ Peer.prototype.onrequest = function (request) {
 
     function onvalue (err, value) {
       if (err) return self.destroy(err)
+      if (!self.stream) return // closed
 
       if (!self.remoteBitfield) self.remoteBitfield = bitfield()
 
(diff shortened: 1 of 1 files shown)


mafintosh:
Missed the PR. Think this is fixed in master. Reopen if not :)

PR #149, opened 2018-03-31
Support onwrite hook

mafintosh:
``` js
var feed = hypercore('./feed', {
  onwrite: function (index, data, cb) {
    console.log('before write', index, data)
    cb()
  }
})

feed.append('hello world') // triggers the hook

// also triggered if data is written during replication
```

Needs docs and tests. Just testing this out for now


pfrazee:
Seems good to me, what was the motivation for this?

mafintosh:
@pfrazee for doing local indexing as data is coming in. think i use this to easily track latest files in hyperdrive

fsteff:
Does this update the data if you pass it to the callback?

mafintosh:
@fsteff we can make it overridable for sure by accepting a buffer in the cb (my usecase was different)

mafintosh:
6.13.0

PR #150, opened 2018-04-24
remove blank alloc

yoshuawuyts:
Found a noop alloc. Not seeing this code being used for anything, so figured it would be okay to remove. Thanks!


mafintosh:
good catch

PR #151, opened 2018-04-25
Enable access to root hashes

### LINKED ISSUE #142: Expose access to root hashes
To provide for https://github.com/mafintosh/hyperdb/issues/41 , it seems an API function must be added to retrieve the root hashes that concisely hash the content of the node.

These hashes are more useful than just the signature already provided, because signatures that only verify a portion of the data are not always shared or available, whereas root hashes are.  For discussion see for example https://botbot.me/freenode/dat/2018-01-20/?msg=95975498 .

I believe this could basically be implemented by exposing `Feed.prototype._getRootsToVerify` .  I imagine the implementation could be the same as `Feed.prototype.verify` but it would not call `crypto.tree` nor `crypto.verify`, rather returning the roots as-is.

Note: last PR I had accepted required spending some time on IRC to draw attention to it, and including tests and documentation updates.

denim2x:
Root hashes can be accessed via Feed::rootHashes; fixes #142

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,12 @@ block associated with the signature.
 Callback is called with `(err, success)` where success is true only if the signature is
 correct.
 
+#### `feed.rootHashes(index, callback)`
+
+Retrieve the root *hashes* for given `index`.
+
+Callback is called with `(err, roots)`; `roots` is an *Array* of root *hashes*.
+
 #### `var number = feed.downloaded([start], [end])`
 
 Returns total number of downloaded blocks within range.
--- a/index.js
+++ b/index.js
@@ -505,7 +505,7 @@ Feed.prototype.signature = function (index, cb) {
 Feed.prototype.verify = function (index, signature, cb) {
   var self = this
 
-  this._getRootsToVerify(index * 2 + 2, {}, [], function (err, roots) {
+  this._withRoots(index, function (err, roots) {
     if (err) return cb(err)
 
     var checksum = crypto.tree(roots)
@@ -516,6 +516,14 @@ Feed.prototype.verify = function (index, signature, cb) {
       cb(null, true)
     }
   })
+}
+
+Feed.prototype.rootHashes = function (index, cb) {
+  this._withRoots(index, cb)
+}
+
+Feed.prototype._withRoots = function (index, cb) {
+  this._getRootsToVerify(index * 2 + 2, {}, [], cb)
 }
 
 Feed.prototype.seek = function (bytes, opts, cb) {
--- a/test/basic.js
+++ b/test/basic.js
@@ -84,6 +84,36 @@ tape('verify', function (t) {
   })
 })
 
+tape('rootHashes', function (t) {
+  t.plan(9)
+
+  var feed = create()
+  var evilfeed = create(feed.key, {secretKey: feed.secretKey})
+
+  feed.append('test', function (err) {
+    t.error(err, 'no error')
+
+    evilfeed.append('t\0st', function (err) {
+      t.error(err, 'no error')
+
+      var result = []
+
+      feed.rootHashes(0, onroots)
+      evilfeed.rootHashes(0, onroots)
+
+      function onroots (err, roots) {
+        t.error(err, 'no error')
+        t.ok(roots instanceof Array)
+        result.push(roots)
+        if (result.length < 2) return
+        t.notEqual(result[0], result[1])
+        t.equal(result[0].length, result[1].length)
+        t.notEqual(Buffer.compare(result[0][0].hash, result[1][0].hash), 0)
+      }
+    })
+  })
+})
+
 tape('pass in secret key', function (t) {
   var keyPair = crypto.keyPair()
   var secretKey = keyPair.secretKey


mafintosh:
Since this is very different than the other verify method I'd much prefer to add thing under a new method called `feed.rootHashes` fx (naming up to you)

mafintosh:
Looks great otherwise

denim2x:
@mafintosh it's done

test/basic.js
```diff
@@ -54,7 +54,7 @@ tape('flush', function (t) {
 })
 
 tape('verify', function (t) {
-  t.plan(9)
+  t.plan(4)
```
xloem: hey @denim2x this change looks like a leftover error from previous work ^_^  I think this should say 9 here?

I noticed this because it's me who wrote the old code that got changed.

test/basic.js
```diff
@@ -84,6 +84,31 @@ tape('verify', function (t) {
...
+      t.error(err, 'no error')
+
+      feed.commend(0, function (err, roots) {
+        t.error(err, 'no error')
+        t.ok(roots instanceof Array)
+      })
+
+      evilfeed.commend(0, function (err, roots) {
+        t.ok(err === null, 'no error')
+        t.ok(roots instanceof Array)
```
xloem: This is good test coverage.  A nitpick would be that the test doesn't check that evilfeed and feed have different root hashes (the point of getting root hashes is to ensure matching content is the same and differing content is different without having to compare all the content).

It would be good to check evilfeed after the roots of feed are returned, and assert that the two have the same number of hashes but that those hashes are different.

README.md
```diff
@@ -184,6 +184,12 @@ block associated with the signature.
 Callback is called with `(err, success)` where success is true only if the signature is
 correct.
 
+#### `feed.commend(index, callback)`
```
xloem: I'm not sure what 'commend' means personally =S  I might call the function 'rootHashes' if nobody else cares !

xloem:
@denim2x I noticed a bug on line 57
Not sure why mafintosh hasn't replied; he seems pretty busy and I had to ask him a lot on irc last time.

README.md
mafintosh: Yes I was just about to comment that, rootHashes is better. I had to lookup commend in the dictionary.

test/basic.js
mafintosh: :+1: on @xloem comments

mafintosh:
@denim2x LGTM on @xloem's comments, sorry for being busy.

denim2x:
thanks for the feedback

index.js
```diff
@@ -518,6 +518,16 @@ Feed.prototype.verify = function (index, signature, cb) {
   })
 }
 
+Feed.prototype.rootHashes = function (index, cb) {
+  this._withRoots(index, function (err, roots) {
```
mafintosh: quick question, why is this not just passing index, cb directly?

index.js
```diff
@@ -518,6 +518,16 @@ Feed.prototype.verify = function (index, signature, cb) {
...
+    if (err) { cb(err) } else { cb(null, roots) }
```
mafintosh:
``` js
if (err) return cb(err)
cb(null, roots)
```

instead, assuming this closure is necessary (see above q)

test/basic.js
```diff
@@ -84,6 +85,42 @@ tape('verify', function (t) {
...
+      t.error(err, 'no error')
+
+      var result = new Assembly()
+      result.on('add', function (ev) {
+        if (result.length === 2) {
+          t.notEqual(result[0], result[1])
+          t.equal(result[0].length, result[1].length)
+          t.notEqual(Buffer.compare(result[0][0].hash, result[1][0].hash), 0)
+        }
+      })
```
mafintosh: this is a bit confusing to read. why not just name the `function (ev)` here and call it after doing result.push below? also removes the need to an additional dep.

``` js
var results = []

feed.rootHashes(0, onroot)
evilfeed.rootHashes(0, onroot)

function onroots (err, roots) {
  t.error(err, 'no error')
  t.ok(roots instanceof Array)
  result.push(roots)
  if (result.length < 2) return
  t.notEqual(result[0], result[1])
  t.equal(result[0].length, result[1].length)
  t.notEqual(Buffer.compare(result[0][0].hash, result[1][0].hash), 0)
}
```

denim2x:
fixed it

mafintosh:
Nice! @denim2x is this an issue? https://github.com/mafintosh/hypercore/pull/151/files#r185249278 other than that LGTM from :tada:

denim2x:
@mafintosh doubtful - since the number assertions in `verifiy` has increased + it's tested and works

xloem:
i’d imagine the test could be terminating early because 4 assertions are planned for in verify, even though 9 are present.  it would still appear to pass but should be fixed.

denim2x:
@mafintosh that's right (I misunderstood @xloem's comment)

index.js
```diff
@@ -505,7 +505,7 @@ Feed.prototype.signature = function (index, cb) {
 Feed.prototype.verify = function (index, signature, cb) {
   var self = this
 
-  this._getRootsToVerify(index * 2 + 2, {}, [], function (err, roots) {
+  this._withRoots(function (err, roots) {
```
mafintosh: Isn't this missing index? a bit confused actually

xloem:
@denim2x I'm sorry but that failure indicates your changes have broken the verify function.  It's important to keep the tests all active; they indicate when something is broken !

denim2x:
the call to `_withRoots` was missing the `index` argument - that's fixed now

mafintosh:
merrrrged

mafintosh:
6.14.0 :tada:

PR #153, opened 2018-05-03
Clarify root hashes documentation.

xloem:
Clarifies that the root hashes implementation does not actually provide a straight array of hashes, but rather the objects which wrap them.


mafintosh:
Nice, thank you.

PR #156, opened 2018-05-21
Make sure the stream is open before using it (close #146)

### LINKED ISSUE #146: Cannot read property 'data' of null in replicate.js

pfrazee:
Resubmitting this PR rebased onto master. We're still getting it (https://github.com/beakerbrowser/beaker/issues/953)

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -63,6 +63,7 @@ function Peer (feed, opts) {
 }
 
 Peer.prototype.onwant = function () {
+  if (!this.stream) return
   // TODO: reply to the actual want context
   this.remoteWant = true
   var rle = this.feed.bitfield.compress()
@@ -119,6 +120,7 @@ Peer.prototype.onrequest = function (request) {
 
     function onvalue (err, value) {
       if (err) return self.destroy(err)
+      if (!self.stream) return // closed
 
       if (!self.remoteBitfield) self.remoteBitfield = bitfield()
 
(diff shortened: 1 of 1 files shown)


mafintosh:
I don't think that bug is happening on latest hypercore. Are you sure that is the case? The line numbers in the beaker screenshot doesn't seem to align with the code base here and we never set stream to null anymore (i think we reverted that).

pfrazee:
Oh hmm. Yeah I'll close then.

Issue #154, opened 2018-05-06
'close' event not emitted


rhodey:
According to the documentation, there should be an event (`close`) that gets emitted *"when the feed has been fully closed"*. It appears to me that this event is not being emitted correctly, heres an example which I ran using `hypercore@6.14.0`:

```javascript
const hypercore = require('hypercore')

function openFeed() {
  const feed = hypercore('./feed', { valueEncoding: 'binary' })
  return new Promise((res, rej) => {
    feed.on('ready', () => res(feed))
    feed.on('error', (err) => {
      console.log('hyper.error', err)
      feed.close()
      rej(err)
    })
    feed.on('close', () => console.log('hyper.close.1'))
  })
}

openFeed().then((feed) => {
  setTimeout(() => {
    console.log('closing...')
    feed.close((err) => console.log('hyper.close.2', err))
  }, 2500)
}).catch((err) => console.log(err))
setTimeout(() => console.log('done'), 12000)
```

## expected result
```
closing...
hyper.close.1
hyper.close.2 null
done
```

## actual result
```
closing...
hyper.close.2 null
done
```

mafintosh:
Yea this seems like a bug for sure

mafintosh:
This should be fixed

PR #157, opened 2018-05-23
emit 'close' event after core has been closed. closed (#154).

rhodey:
Emit `close` event after core has been closed. Maybe we don't want to emit `close` if the close callback is invoked with an error, I'm not sure. In my situation I think I prefer emitting the event even if close failed, because close will probably just keep failing again, and I'm more interested in "when is is OK to exit".


index.js
```diff
@@ -1001,7 +1001,10 @@ Feed.prototype.close = function (cb) {
   this._ready(function () {
     self.writable = false
     self.readable = false
-    self._storage.close(cb)
+    self._storage.close(function (err) {
+      self.emit('close')
```
mafintosh: should only be emitted if no error occured during storage.close. Also mind adding a flag '.closed` or similar so we don't emit this twice?

mafintosh:
@rhodey nice catch! take a look at my comment.

rhodey:
@mafintosh thanks for the comments :) what do you think about my `.closed` logic? do we wanna just prevent the emitting of `> 1` event or should we no-op the whole method if it has already been called?

mafintosh:
This looks good to me

mafintosh:
@rhodey oh mind adding a test? then we can't a regression in the future

mafintosh:
6.15.0

PR #163, opened 2018-06-25
impl inspect.custom

mafintosh:
Makes `console.log(feed)` print something similar to:

```
Hypercore(
  key: dc3b17..ed
  discoveryKey: 09ba31..48
  opened: true
  writable: true
  length: 69
  byteLength: 1196
  peers: 0
)
```


pfrazee:
👍 good stuff

PR #167, opened 2018-07-08
sync-bitfield event to know when the bitfield has been written

substack:
For [multifeed-index](https://github.com/noffle/multifeed-index), I need an event to know when I can read records from the log. The `'append'` event happens before the records are written and the `'sync'` event happens before the bitfield is updated after replication. With this additional event, I can tail a hypercore log without subtle timing bugs around replication.

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,6 +346,10 @@ Emitted when the feed has been appended to (i.e. has a new length / byteLength)
 
 Emitted every time ALL data from `0` to `feed.length` has been downloaded.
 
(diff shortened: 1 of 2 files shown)


noffle:
Nice!

Could `"written"` or `"appended"` also be a valid name choices? It's not as obvious to me, without deep knowledge of hypercore internals, that `"sync-bitfield"` means that an entry has been fully written.

cc @mafintosh

pfrazee:
Why was this closed?

PR #168, opened 2018-07-19
Start download after changing start index when using tail

jimpick:
Otherwise all blocks get downloaded.

(still needs a test)


martinheidegger:
A test for it would be nice!

mafintosh:

PR #170, opened 2018-08-20
much faster block selection

mafintosh:
Instead of walking the remote bitfield to find a fitting chunk this now *once* produces

```
bitfield = remote-blocks & !local-blocks
```

Then to select a block this bitfield is simply iterated (very fast, log32(n))

@pfrazee can you sanity check this for silly errors? it's v critical code


mafintosh:
(@pfrazee tested this on irc for me with success)

PR #173, opened 2018-09-16
Handle case for `download()` where no peers are present.

RangerMauve:
Should fix `download()` stalling out if you have no peers but have the content locally already.

Related to https://github.com/beakerbrowser/beaker/issues/1153 and https://github.com/tswast/dat-pin-box/issues/10 (cc @pfrazee )

Some questions:

- How should I test that this works?
- Should we throw an error if there are no peers and it isn't downloaded?

### DIFF
--- a/index.js
+++ b/index.js
@@ -325,15 +325,19 @@ Feed.prototype.download = function (range, cb) {
   if (!cb) cb = noop
   if (!this.readable) return cb(new Error('Feed is closed'))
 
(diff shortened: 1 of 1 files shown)


mafintosh:
@RangerMauve 

1. You should be able to test this by having one feed you append data to and the tell it to download the same data you just appended
1. No don't throw. Current behaivor looks good (same a .get)

Other that missing test LGTM

mafintosh:
Fixed in latest

PR #176, opened 2018-10-13
append passes seq to the callback. fixes #94

### LINKED ISSUE #94: `.append()` should invoke `callback(null, index)`
`.append()`'s callback should give the index to the caller.

mafintosh:
It's surprisingly tricky to get the seq at which a piece of data was appended.

This PR adds it the append callback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,7 +237,7 @@ False otherwise.
 
 Append a block of data to the feed.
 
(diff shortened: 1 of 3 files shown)


mafintosh:
(gonna add some tests for this as well)

mafintosh:
6.20.0

PR #181, opened 2018-11-12
Tweak order in api docs

ralphtheninja:
@mafintosh I'm not sure you agree or if you had some other idea in mind when writing the docs, this is just a suggestion from me :)

* move .append() to the top, it's an append only log and this method should be the first, because this is most likely the first method to be used, also close to .get()
* move the properties down closer to the events (they all refer to the 'ready' event), also they are probably not the first things you want to read about


karissa:
lgtm!

mafintosh:
Thanks!

Issue #183, opened 2018-11-25
Signatures not getting sent with data by Beaker


pfrazee:
Beaker isn't sending a signature with its proofs. I'm still isolating why. Here's what I've got so far:

 - `Feed.prototype.proof()` is calculating `needsSig` to be `false`. This causes Hypercore to avoid fetching the signature.
 - `needsSig` is `false` because `TreeIndex.prototype.proof()` is returning `verifiedBy: 0`
 - `TreeIndex.prototype.proof()` is returning `verifiedBy: 0` because it reaches https://github.com/mafintosh/hypercore/blob/master/lib/tree-index.js#L67

I don't totally understand the logic of https://github.com/mafintosh/hypercore/blob/master/lib/tree-index.js#L53-L65 so I can't trace the cause further. I ran `audit()` on the metadata and content hypercores and got no invalid blocks.

There's a pretty good indication that this is affecting multiple people. Fritter went quiet!

LMK what else you need.

mafintosh:
Should be fixed now

PR #184, opened 2018-12-10
Reduce bitfield footprint

mafintosh:
Turns the memory footprint constant (low) on low sparse clients

### DIFF
--- a/test/replicate.js
+++ b/test/replicate.js
@@ -637,14 +637,16 @@ tape('replicate with onwrite', function (t) {
   })
 })
 
-tape.skip('replicate from very sparse', function (t) {
+tape('replicate from very sparse', function (t) {
   t.plan(3)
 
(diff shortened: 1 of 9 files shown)


test/replicate.js
```diff
@@ -637,7 +637,7 @@ tape('replicate with onwrite', function (t) {
   })
 })
 
-tape('replicate from very sparse', function (t) {
+tape.skip('replicate from very sparse', function (t) {
```
mafintosh: need to investigate why this is slower before merging

mafintosh:
This also implements "only subscribe to specific want regions" which makes hypercore scale to basically any size of stream

mafintosh:
@pfrazee would you say this fixes #183 ?

mafintosh:
6.22.1

PR #187, opened 2018-12-18
Removed dependency on buffer-equals as it has been deprecated.

martinheidegger:
`buffer-equals` has been deprecated with the reference to use `Buffer.equals` instead. This PR does just that.


mafintosh:
:+1: should we merge your other PR first?

PR #188, opened 2018-12-18
Fix: timing error that resulted in missing remoteId on peer when peer is added.

martinheidegger:
For some reason the `remote-add` event was submitted before the handshake. This is problematic as the remoteId doesn't exist at that point. This PR adds a test to check for the remoteId and fixes the test by adding the peer after the handshake.

I am not sure this is the best approach to this, but all the tests pass.


martinheidegger:
I just added commit 41ffa91 that comes with a sturdier solution:

Handshake may come often or never, depending in which state the stream is passed in. This means that `.ready()` should be triggered if a `remoteId` is available and only the first time that a handshake is submitted.
Multiple handshakes can technically update the remoteId. If the remoteId changes on update it is a good idea sending out a `remote-update` to identify the change in remote identification.

Not sure how to best explicitly test for those cases.

lib/replicate.js
```diff
@@ -28,13 +29,29 @@ function replicate (feed, opts) {
...
+        peer.feed.emit('remote-update', peer)
+      }
     })
+    var firstTime = true
 
-    // stream might get destroyed on feed init in case of conf errors
-    if (stream.destroyed) return
+    triggerReady()
+
+    function triggerReady () {
```
mafintosh: Can you make this return false as well instead of undefined?

index.js
```diff
@@ -22,6 +22,7 @@ var bufferFrom = require('buffer-from')
 var bufferAlloc = require('buffer-alloc-unsafe')
 var inspect = require('inspect-custom-symbol')
 var pretty = require('pretty-hash')
+var safeBufferEquals = require('./lib/saveBufferEquals')
```
mafintosh: rename the file to lib/safe-buffer-equals.js

PR #189, opened 2019-01-17
Specify key type (avoid browser Buffer/ArrayBuffer gotcha)

aral:
hypercore constructor accepts Node.js Buffer instances, not ArrayBuffer instances. This is a gotcha when using hypercore in the browser. This change explicitly documents this behaviour.


martinheidegger:
I find that note very helpful. hypercore uses internally [buffer-from](https://www.npmjs.com/package/buffer-from) to convert a string-input into a buffer key. Should be helpful in the browser as well.

mafintosh:
Would be cool to support uint8arrays plus arraybuffers as well in the future (shouldn't be too hard) :+1:

PR #190, opened 2019-01-17
Add reminder that the onwrite handler has to call cb()

aral:
This was not immediately clear to me and my appends were not working without it.


martinheidegger:
Maybe also rename the argument to `next`?

aral:
@martinheidegger :+1: `next` would make it clearer that this works like middleware. I would not have been confused if that’s what it had been called in the docs.

PR #192, opened 2019-01-22
batch all writes to storage in append

mafintosh:
Makes all appends to O(log(n)) writes to storage instead of O(n)


mafintosh:
Yea, we should make a random-access-storage wrapper that just counts writes and reads for this.

PR #199, opened 2019-01-25
Add bitfield lock

mafintosh:
Uses fd-lock (optional dep as it is native with prebuilds) to lock access to the bitfield. Avoids accidental forks by running two hypercores next to each other accessing the same data.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1363,7 +1363,10 @@ function isBlock (index) {
 
 function defaultStorage (dir) {
   return function (name) {
-    return raf(name, {directory: dir})
+    try {
+      var lock = name === 'bitfield' ? require('fd-lock') : null
+    } catch (err) {}
+    return raf(name, {directory: dir, lock: lock})
   }
 }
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1363,7 +1363,10 @@ function isBlock (index) {
 
 function defaultStorage (dir) {
   return function (name) {
-    return raf(name, {directory: dir})
+    try {
+      var lock = name === 'bitfield' ? require('fd-lock') : null
+    } catch (err) {}
```
yoshuawuyts: Perhaps a call to `debug()` might be helpful here, in case this ever needs to be debugged.

yoshuawuyts:
Perhaps also set the minimum version of `random-access-file` to [2.1.0](https://github.com/random-access-storage/random-access-file/commits/master) to prevent any caching problems when upgrading.

mafintosh:
Fixed the min version. We don't use debug anywhere (actually not a fan of debug, *runs away*), so adding it here only is a bit premature.

yoshuawuyts:
@mafintosh haha, yeah fair. Was more thinking that the (old) networking layer uses debug a lot, and it's helped us out more than once when building dat-desktop. Figured it might be helpful in this case too in case there's ever lock contention problems. But shrug, it's not super important.

Issue #196, opened 2019-01-24
test that all storage drivers get closed on .close()


mafintosh:
(random-access-* have a .closed prop for this)

PR #201, opened 2019-01-25
Testing reduced reduce writes

martinheidegger:
As hinted on in #196 this PR adds a test to see that multiple writes in close succession. I decided to not just look at the number called but also at the data transported. Using the stack it might be easier to identify in an error case what went wrong.

### DIFF
--- a/test/basic.js
+++ b/test/basic.js
@@ -1,4 +1,5 @@
 var create = require('./helpers/create')
+var createTrackingRam = require('./helpers/create-tracking-ram')
 var crypto = require('hypercore-crypto')
 var tape = require('tape')
 var hypercore = require('../')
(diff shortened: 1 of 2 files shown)


test/basic.js
```diff
@@ -404,3 +405,22 @@ tape('append and createWriteStreams preserve seq', function (t) {
...
+  var feed = hypercore(trackingRam)
+  var ws = feed.createWriteStream()
+
+  ws.write('ab')
+  ws.write('cd')
+  ws.write('ef')
+  ws.end(function () {
+    t.deepEquals(trackingRam.log.data, [
+      { write: [ 0, Buffer.from('ab') ] },
+      { write: [ 2, Buffer.from('bcef') ] }
```
mafintosh: I think this should be cdef

martinheidegger: Totally, ... fixed now _(memo: sleep more)_

mafintosh:
woot

PR #202, opened 2019-02-05
Pluggable async crypto

emilbayes:
Main motivation is being able to use HSMs or Yubi keys to sign hypercores


emilbayes:
Added docs now. Obviously there is still work to be done in replication so other peers know which suite of crypto algorithms are used

mafintosh:
Will merge and release later when I’m on my laptop

mafintosh:
7.2.0

PR #205, opened 2019-03-07
Fix link to Dat paper in README

HerbCaudill:
This URL is no longer valid: https://github.com/datproject/docs/blob/master/papers/dat-paper.pdf

Found the paper here: https://github.com/datprotocol/whitepaper/blob/master/dat-paper.pdf


HerbCaudill:
Reading the paper now 😂

mafintosh:
Hahahaha, thanks for the fix!

PR #206, opened 2019-03-23
Run tests on Windows

gmaclennan:
This PR runs tests on Windows in Travis. It switches to using random-access-file in tests rather than random-access-memory, in order to better replicate a real-world scenario in tests and surface any peculiarities of the filesystem (which it did).

Tests are inconsistently failing on Windows. Help is most welcome figuring out why this is failing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,7 @@ node_js:
   - '6'
   - '8'
   - '10'
+os:
+  - windows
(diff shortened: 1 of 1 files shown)


gmaclennan:
PR updated to remove switch to random-access-file (see #209). This PR now only updates travis to run tests on Windows.

PR #209, opened 2019-03-29
Test using random-access-file not random-access-memory

gmaclennan:
Switch to using random-access-file in tests, in order to test in an environment more similar to actual use and catch unusual filesystem related bugs that can crop up on Windows etc.


mafintosh:
Prefer to keep it memory only in the test cases, but down to change my mind if we do it dead simply

PR #210, opened 2019-03-30
test filesystem compatibility, esp for windows

okdistribute:
Switch to using random-access-file in tests, in order to test in an environment more similar to actual use and catch unusual filesystem related bugs that can crop up on Windows etc.

This also cleans up files after creating them, superseding https://github.com/mafintosh/hypercore/pull/209

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,11 +35,13 @@
   "devDependencies": {
     "nyc": "^13.1.0",
     "random-access-memory": "^3.1.0",
+    "rimraf": "^2.6.3",
     "shuffle-array": "^1.0.1",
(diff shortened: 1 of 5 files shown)


karissa:
@martinhdiedegger that's already the case. The issue here is that the tests use random access memory so filesystem differences are not being tested

https://github.com/mafintosh/hypercore/blob/master/.travis.yml

martinheidegger:
Oh that happened in #206 (quite recently) 👍

mafintosh:
We have fixutres in there now to test compat

PR #211, opened 2019-04-15
Add network stats

andrewosh:
This PR adds the `stats` property, which returns a handful of (currently only network-related) statistics. Stats collection can be disabled by settings `opts.stats` to `false` when creating the hypercore.

Currently, network stats are tracked both a) in total, and b) for each connected peer. Historical stats for disconnected peers are not stored.

The returned stats object has the following form (for a source feed that's uploaded a single 100 byte block to a currently-connected peer):
```js
{
  totals: {
    uploadedBytes: 100,
    uploadedBlocks: 1,
    downloadedBytes: 0,
    downloadedBlocks: 0
  },
  peers: [
    {
      uploadedBytes: 100,
      uploadedBlocks: 1,
      downloadedBytes: 0,
      downloadedBlocks: 0
    },
    ...
  ]
}
```

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -86,6 +86,8 @@ function Peer (feed, opts) {
   this._defaultDownloading = this.downloading
   this._iterator = this.remoteBitfield.iterator()
 
+  this._stats = null
+
   if (opts.stats) {
     this._stats = {
       uploadedBytes: 0,
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -127,7 +127,7 @@ Peer.prototype.ondata = function (data) {
   this.feed._putBuffer(data.index, data.value, data, this, function (err) {
     if (err) return self.destroy(err)
     if (data.value) self.remoteBitfield.set(data.index, false)
-    if (self._stats) {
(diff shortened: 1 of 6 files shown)


lib/replicate.js
```diff
@@ -85,6 +85,15 @@ function Peer (feed, opts) {
   this._destroyed = false
   this._defaultDownloading = this.downloading
   this._iterator = this.remoteBitfield.iterator()
+
+  if (opts.stats) {
+    this._stats = {
```
mafintosh: this._stats = opts.stats ? null : { ... }

andrewosh: :+1:

joehand:
Awesome!

Looks like these stats are only *per session*, i.e. doesn't count already downloaded data. This will be nice for some UI info, but may cause some confusion trying to calculate total progress information. I wonder if the API should be more specific, e.g. `sessionStats` or `networkStats`, or also include the downloaded stats.

mafintosh:
@joehand persisted stats are prob out of scope here (at least for now), but yea sessionStats is not a bad name and we might wanna alias it (already released)

PR #213, opened 2019-06-23
expand crypto option

lukeburns:
Not sure this is of interest, but I wanted to try plugging in [`wasm-crypto`](https://github.com/jedisct1/wasm-crypto) in place of `hypercore-crypto` to use signatures over the Ristretto group. This PR makes it possible to specify a keyPair function that hypercore will use to generate keys, along with the size of keys/signatures. [lukeburns/wasm-crypto](https://github.com/lukeburns/wasm-crypto) has an api matching that of `hypercore-crypto` that can be used in the following example.

```js
var hypercore = require('hypercore')
var { keyPair, sign, verify } = require('wasm-crypto')

var crypto = {
  keyPair,
  sign (data, sk, cb) {
    return cb(null, sign(data, sk))
  },
  verify (data, sig, pk, cb) {
    return cb(null, verify(data, sig, pk))
  },
  secretKeySize: 32, // 32-byte secret keys
  publicKeySize: 32,
  signatureSize: 64
}

var feed = hypercore('./tmp', { valueEncoding: 'json', crypto })

feed.append({
  hello: 'world'
})

feed.append({
  hej: 'verden'
})

feed.append({
  hola: 'mundo'
})

feed.flush(function () {
  console.log('Appended 3 more blocks, %d in total (%d bytes)\n', feed.length, feed.byteLength)

  feed.createReadStream()
    .on('data', console.log)
    .on('end', console.log.bind(console, '\n(end)'))
})
```


lib/storage.js
```diff
@@ -257,18 +265,16 @@ Storage.prototype.open = function (opts, cb) {
     })
   })
 
-  this.signatures.write(0, header(1, 64, 'Ed25519'), done)
+  this.signatures.write(0, header(1, this.signatureSize, 'Ed25519'), done)
```
emilbayes: Does the const string here also need changing? Ie. are the ristretto255 points "decompressed" back into valid ed25519 points?

index.js
```diff
@@ -22,12 +22,13 @@ var safeBufferEquals = require('./lib/safe-buffer-equals')
...
 var defaultCrypto = {
+  keyPair: crypto.keyPair,
   sign (data, sk, cb) {
     return cb(null, crypto.sign(data, sk))
   },
 
-  verify (sig, data, pk, cb) {
-    return cb(null, crypto.verify(sig, data, pk))
+  verify (data, sig, pk, cb) {
+    return cb(null, crypto.verify(data, sig, pk))
```
emilbayes: Was the swapping of args here intentional?

lukeburns: Yes, to match the api of hypercore-crypto

lib/storage.js
lukeburns: Ah yes we should change the string too. Does the protocol use this to ensure peers are using the same kind of signatures?

emilbayes: Not yet, but we’re working on updating the handshake to communicate info like this

lukeburns: Added signatureType

mafintosh:

PR #214, opened 2019-06-26
Surfaced protocol extensions

RangerMauve:
This PR adds first-party support for [`extension`](https://github.com/mafintosh/hypercore-protocol/blob/master/schema.proto#L15) messages.
It adds methods for sending extension messages either to all peers, or a specific peer, and receiving extension messages from peers.
This should make things like [multifeed](https://github.com/kappa-db/multifeed) easier to write.

Added:

- `extensions` option in constructor
- `extensions` property on hypercore
- Pass `extensions` option to replication if one doesn't exist
- `extension` method to Peers for sending extension message to a peer
- `onextension` method in peer gets [invoked as an event listener](https://github.com/mafintosh/hypercore-protocol/blob/master/feed.js#L151)
- `extension` method on hypercore which sends the extension to all peers
- `'extension'` event on hypercore that gets emitted along with the peer that sent it

TODO:

- [x] Tests
- [x] Docs


mafintosh:
@RangerMauve hate to be a nit, but can you remove the standard --fix one? that should be a new PR that upgrades standard in the repo.

Do ./node_modules/.bin/standard --fix to use local one

RangerMauve:
Sorry about that! Fixed it. :)

PR #215, opened 2019-07-05
ACK feature

substack:
This PR is based on https://github.com/mafintosh/hypercore/pull/130 but adds:

* updated to the latest hypercore
* adds many tests
* uses an `'ack'` event on the replication stream instead of emitting events through the peer instance (only available by listening on `feed.on('peer-add', fn)` and the peer instance isn't an EventEmitter in mainline anyways)
* sets the length in an acknowledgement HAVE to 0

The original technique in the ACK PR mostly works but creates a bit of extra noise with false positives. This extra noise would probably not be too much of an issue in practice for many use cases because any ACK'd records would also be present on the remote machine (if the remote is being honest). But setting the length to 0 for ACKs avoids this noise. A better option might be to add a special type for ACK messages, but this patch works with that is currently available in hypercore-protocol.

There are some legacy cases for length=0 in HAVEs but the ACK cases are narrow and only reachable when ACK is enabled and a bitfield is not set on the HAVE, which the legacy branch seems to expect.

If this patch is accepted it would be good to document the convention it depends on, where a HAVE with length=0 and no bitfield is considered an ACK. This also means that ACKs are limited to acknowledging individual sequences, not whole ranges at a time (since the length field is set to zero). Negative values could be used, but at that point I think it would be better to have a dedicated ACK message type.

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -30,7 +30,6 @@ function replicate (feed, opts) {
     var peer = new Peer(feed, opts)
     peer.feed = feed
     peer.stream = stream.feed(feed.key, {peer: peer})
-    peer._replicateStream = stream
 
     peer.remoteId = stream.remoteId
     stream.setMaxListeners(0)
@@ -247,7 +246,7 @@ Peer.prototype.ontick = function () {
 
 Peer.prototype.onhave = function (have) {
   if (this.ack && have.ack && !have.bitfield && this.feed.bitfield.get(have.start)) {
-    this._replicateStream.emit('ack', have.start)
+    this.stream.stream.emit('ack', have)
     return
   }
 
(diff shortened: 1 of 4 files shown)


substack:
I think this PR would best be redone to add an optional boolean ack field to HAVE messages in hypercore protocol instead of setting length=0 on HAVEs. This would be less hacky and would also allow for ACKs over ranges of sequences.

substack:
Updated this patch to depend on https://github.com/mafintosh/hypercore-protocol/pull/37 so the CI ought to fail until that gets merged upstream.

lib/replicate.js
```diff
@@ -30,6 +30,7 @@ function replicate (feed, opts) {
     var peer = new Peer(feed, opts)
     peer.feed = feed
     peer.stream = stream.feed(feed.key, {peer: peer})
+    peer._replicateStream = stream
```
mafintosh: this is stored at peer.stream.stream already (could be named better yes)

mafintosh:
7.6.0

PR #217, opened 2019-07-20
Update README.md

perguth:
Some changes here and there for consistency, clarity and readability. ✨

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,7 +340,7 @@ Populated after `ready` has been emitted. Will be `false` before the event.
 
 #### `feed.readable`
 
-Can we read from this feed? After closing a feed this will be false.
+Can we read from this feed? After closing the feed this will be false.
 
 Populated after `ready` has been emitted. Will be `false` before the event.
 
--- a/README.md
+++ b/README.md
@@ -201,8 +201,8 @@ Retrieve the root *hashes* for given `index`.
 Callback is called with `(err, roots)`; `roots` is an *Array* of *Node* objects:
 ```
 Node {
-  index: location in the merkle tree of this root
-  size: total bytes in children of this root
+  index: location in the merkle tree of this root,
+  size: total bytes in children of this root,
   hash: hash of the children of this root (32-byte buffer)
 }
 ```
--- a/README.md
+++ b/README.md
@@ -414,11 +414,11 @@ Emitted when a data block has been downloaded.
 
 #### `feed.on('upload', index, data)`
 
-Emitted when a data block is uploaded.
+Emitted when a data block has been uploaded.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -414,11 +414,11 @@ Emitted when a data block has been downloaded.
 
 #### `feed.on('upload', index, data)`
 
-Emitted when a data block is uploaded.
+Emitted when a data block has been uploaded.
```
perguth: I am guessing that though.

README.md
```diff
@@ -134,7 +134,7 @@ Callback is called with `(err, data)`
 
 #### `feed.getBatch(start, end, [options], callback)`
 
-Get a range of blocks efficiently. Options include
+Get a range of blocks efficiently. End index is non-inclusive. Options include
```
emilbayes: Another word for "non-inclusive" could be "exclusive"

README.md
```diff
@@ -352,7 +352,7 @@ Populated after `ready` has been emitted. Will be `null` before the event.
 
 #### `feed.discoveryKey`
 
-Buffer containing a key derived from the feed.key.
+Buffer containing a key derived from the feeds public key.
```
emilbayes: I think feeds should have an apostrophe on the end to be grammatically correct, as it is possessive. Ie. `from the feeds' public key`

README.md
emilbayes: I actually think the former is correct as it happens *before* the block is sent over the wire. See https://github.com/mafintosh/hypercore/blob/5a46735bb95e96265dea32f2919a94da120b00fa/lib/replicate.js#L181-L195

emilbayes:
Looks good! Added a few comments :)

README.md
perguth: It's strange but somehow I feel a novice would understand `non-inclusive` more easily. But As you say.

emilbayes: @perguth I don't have any power here, in math/CS I feel like intervals are usually called inclusive and exclusive

README.md
perguth: Ok, clarified. Ty!

README.md
perguth: Ok, I would stick to non-inclusive. It's also used a couple of lines down in this paragraph: https://github.com/mafintosh/hypercore/blob/5a46735bb95e96265dea32f2919a94da120b00fa/README.md#feeddownloadrange-callback

perguth:
Good for merge then. Thanks everyone! 🚀

perguth:
Ping. 🏓

perguth:
thanks 😊

PR #220, opened 2019-08-06
Fixed #219 opts.expectedFeeds no longer ignored when opts.stream is present

### LINKED ISSUE #219: Bug `opts.expectedFeeds` ignored when stream present
In [`lib/replicate.js:15`](https://github.com/mafintosh/hypercore/blob/master/lib/replicate.js#L15) we have the following code:
```js
 if (!stream) {
    if (!opts.expectedFeeds) opts.expectedFeeds = 1
    if (!opts.id) opts.id = feed.id
    stream = protocol(opts)
  }
```
If a stream is not provided, then `expectedFeeds` is optionally lazy-initialized and the opts are passed onto `hypercore-protocol`

I'm attempting to replicate a hyperdrive by passing a stream:

```js
drive.replicate({ stream: myStream })
```
Hyperdrive properly reports that it's a composite-core consisting of [2 feeds](https://github.com/mafintosh/hyperdrive/blob/v10.0.0-rc2/index.js#L674).

But when `opts.stream` was passed, then `opts.expectedFeeds` is completely ignored which causes hyperdrive to prematurely finish replication before the `content` core has been replicated.

I believe this is a bug and that a second condition should be added:
```js
 if (!stream) {
    if (!opts.expectedFeeds) opts.expectedFeeds = 1
    if (!opts.id) opts.id = feed.id
    stream = protocol(opts)
  } else if (opts.expectedFeeds) {   // Fix
    stream.expectedFeeds += opts.expectedFeeds
  }
```
Can someone verify?

telamon:


telamon:
This was superseeded by `hypercore-protocol:v7` `new Protocol(true).prefinalize` object in the form of a [nano-guard](https://github.com/mafintosh/nanoguard)

PR #227, opened 2019-09-12
Companion PR to hypercore-protocol v7 PR

mafintosh:
Tests parsing :) Needs a few doc updates on the updated interface for the replication stream

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -1,4 +1,5 @@
 var Protocol = require('hypercore-protocol')
+var timeout = require('timeout-refresh')
 var bitfield = require('fast-bitfield')
 var set = require('unordered-set')
 var rle = require('bitfield-rle').align(4)
@@ -27,6 +28,11 @@ function replicate (feed, initiator, opts) {
 
     if (err) return stream.destroy(err)
     if (stream.destroyed) return
+
+    if (stream.remoteOpened(feed.key) && !stream.remoteVerified(feed.key)) {
+      stream.close(feed.key)
+      return
+    }
 
     var peer = new Peer(feed, opts)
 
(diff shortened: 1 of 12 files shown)


mafintosh:
Reminder to myself that it needs request timeout logic (and a test for that apparently)

lib/replicate.js
```diff
@@ -345,11 +317,11 @@ Peer.prototype.oncancel = function () {
 }
 
 Peer.prototype.onclose = function () {
-  this.destroy()
+  this._close()
 }
 
 Peer.prototype.have = function (have) { // called by feed
-  if (this.stream && this.remoteWant) this.stream.have(have)
+  if (this.stream && this.remoteWant && this._remoteOpened) this.stream.have(have)
```
andrewosh: When would `this.remoteWant` be true when `this._remoteOpened` isn't? I'd imagine that all wants would have to be received after the first open right?

mafintosh: good point!

test/extensions.js
```diff
@@ -39,10 +39,12 @@ tape('send and receive extension messages', function (t) {
       t.equal(message.toString('hex'), EXAMPLE_MESSAGE.toString('hex'))
     })
 
-    feed1.on('peer-add', function () {
-      feed1.extension(EXAMPLE_TYPE, EXAMPLE_MESSAGE)
+    feed1.once('peer-add', function () {
+      setImmediate(function () {
```
andrewosh: Why does the `extension` call need to be `setImmediate`d here?

mafintosh: It's because the peer-add event is trigger as soon as the peer is added (ie feed.replicate() tick) but it's not setup yet. we prop want to add a peer-ready event for this in general

mafintosh:
Request timeout logic added, including a test. This is good to go on green ci 🚢

PR #230, opened 2019-10-07
add setDownloading setUploading

mafintosh:
allow you to configure data upload/download async

### DIFF
--- a/index.js
+++ b/index.js
@@ -70,6 +70,8 @@ function Feed (createStorage, key, opts) {
   this.readable = true
   this.opened = false
   this.closed = false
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -212,6 +215,25 @@ Feed.prototype.replicate = function (initiator, opts) {
   return stream
 }
 
+Feed.prototype.setDownloading = function (downloading) {
+  this.downloading = downloading
+  this._downloadingSet = true
```
andrewosh: If downloading is `false` should this still be set to `true`?

mafintosh: yes, just whether it's been set or not

PR #231, opened 2019-10-11
spec out new extension api

mafintosh:


README.md
```diff
@@ -429,9 +428,27 @@ The returned object is of the form:
...
+{
+  encoding: 'json' | 'binary' | 'utf-8' | anyAbstractEncoding
+  onmessage (message, peer) {
+    // called when a message is received from a peer
+    // will be decoded using the encoding you provide
+  }
+}
+```
+
+#### `ext.send(message, peer)`
```
andrewosh: Is the `peer` param a `Peer` object or is it an ID? Maybe describe that `peer` is the object emitted on the `peer` event.

index.js
```diff
@@ -208,13 +206,90 @@ Feed.prototype.replicate = function (initiator, opts) {
...
+    if (!encoding) {
+      this.handlers.onmessage(buffer, peer)
+      return
+    }
+
+    let message = null
+
+    try {
+      message = encoding.decode(buffer)
+    } catch (_) {
```
andrewosh: Should extension errors be propagated?

lib/replicate.js
```diff
@@ -46,6 +46,8 @@ function replicate (feed, initiator, opts) {
 }
 
 function Peer (feed, opts) {
+  if (opts.extensions) throw new Error('Per peer extensions is deprecated. Use feed.registerExtension instead')
```
andrewosh: If it's just deprecated it shouldn't throw an error. Update error message to say "not supported"

lib/replicate.js
```diff
@@ -201,6 +202,15 @@ Peer.prototype.onrequest = function (request) {
   }
 }
 
+Peer.prototype._updateOptions = function () {
+  if (this.ack || (this.feed.extensions.length)) {
```
andrewosh: What does an options update have to do with the `ack` flag?

mafintosh: It needs to send an options message if ack is enabled OR you are using extensions, othersise it is not needed

okdistribute: Seems like ack could be implemented as an extension, although might be more work than it's worth

mafintosh: Yea, we should consider that down the line I think

PR #232, opened 2019-10-28
(Bug) Update without hash:false should not download

andrewosh:
Here's a test case where `update({ ifAvailable: true })` either a) downloads the first block or b) does something to the internal state such that `downloaded()` returns 1.


mafintosh:
Adjusted, added, and fixed this test in 8.2.6

PR #233, opened 2019-11-04
be a nanoresource for easier state management

mafintosh:
Should fix some bugs with close not working in some edgecases


mafintosh:
8.3.0

PR #240, opened 2019-12-13
Add upload option to replicate

sammacbeth:
The `upload` option to replicate exists in older docs. This PR brings it back. When set to true data from the feed will not be uploaded to peers.


mafintosh:
Awesome

mafintosh:
8.4.0

PR #245, opened 2020-02-03
Allow passing custom cache as an option

andrewosh:
This PR adds support for a `cache` option of the form:
```js
{
  tree: (number or cache object),
  data: (number or cache object)
}
```

If the `tree` option is either unspecified or is a number, then a HypercoreCache is created with the default byte size 65536 * 40 (the size of 65536 nodes).

If the `data` option is unspecified, then data caching is disabled. If it is a number, then a HypercoreCache is created of the specified byte size.

By allowing the user to inject a cache into Hypercore, one can reuse a single LRU cache across multiple Hypercore instances, improving resource management.

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,7 +99,8 @@ function Feed (createStorage, key, opts) {
   this._overwrite = !!opts.overwrite
   this._storeSecretKey = opts.storeSecretKey !== false
   this._merkle = null
-  this._storage = storage(createStorage, opts.storageCacheSize)
+
+  this._storage = storage(createStorage, opts)
   this._batch = batcher(this._onwrite ? workHook : work)
 
   this._seq = 0
--- a/index.js
+++ b/index.js
@@ -99,7 +99,6 @@ function Feed (createStorage, key, opts) {
   this._overwrite = !!opts.overwrite
   this._storeSecretKey = opts.storeSecretKey !== false
   this._merkle = null
-
   this._storage = storage(createStorage, opts)
   this._batch = batcher(this._onwrite ? workHook : work)
 
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -3,9 +3,11 @@ const HypercoreCache = require('hypercore-cache')
 const DEFAULT_TREE_CACHE_SIZE = 65536 * 40
 
 function createCache (opts) {
+  if (opts.cache === false) return {}
+
   const cacheOpts = opts.cache || {}
   if (cacheOpts.tree === undefined || typeof cacheOpts.tree === 'number') {
-    let cacheSize = cacheOpts.tree || opts.storageCacheSize
+    const cacheSize = cacheOpts.tree || opts.storageCacheSize
     cacheOpts.tree = new HypercoreCache({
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -99,7 +99,8 @@ function Feed (createStorage, key, opts) {
   this._overwrite = !!opts.overwrite
   this._storeSecretKey = opts.storeSecretKey !== false
   this._merkle = null
-  this._storage = storage(createStorage, opts.storageCacheSize)
+
```
mafintosh: remove this whitespace

lib/cache.js
```diff
@@ -0,0 +1,26 @@
+var HypercoreCache = require('hypercore-cache')
```
mafintosh: const

lib/cache.js
```diff
@@ -0,0 +1,26 @@
...
+  if (typeof cacheOpts.data === 'number') {
+    cacheOpts.data = new HypercoreCache({
+      maxByteSize: cacheOpts.data,
+      estimateSize: buf => buf.length
+    })
+  }
+  return cacheOpts
+}
+
+module.exports = {
```
mafintosh: module.exports = createCache

lib/storage.js
```diff
@@ -1,16 +1,17 @@
 var uint64be = require('uint64be')
 var flat = require('flat-tree')
-var alru = require('array-lru')
+
```
mafintosh: remove whitespace

lib/cache.js
```diff
@@ -0,0 +1,26 @@
+var HypercoreCache = require('hypercore-cache')
+
+const DEFAULT_TREE_CACHE_SIZE = 65536 * 40
+
+function createCache (opts) {
+  var cacheOpts = opts.cache || {}
```
mafintosh: const

lib/cache.js
```diff
@@ -0,0 +1,26 @@
...
+  if (cacheOpts.tree === undefined || typeof cacheOpts.tree === 'number') {
+    let cacheSize = cacheOpts.tree || opts.storageCacheSize
```
mafintosh: const

lib/storage.js
```diff
@@ -145,10 +154,8 @@ Storage.prototype.getDataBatch = function (start, n, cb) {
 }
 
 Storage.prototype.getNode = function (index, cb) {
-  if (this.cache) {
-    var cached = this.cache.get(index)
-    if (cached) return cb(null, cached)
-  }
+  var cached = this.cache.tree.get(index)
```
mafintosh: cache is always defined now?

mafintosh: can be null no?

test/cache.js
```diff
@@ -0,0 +1,77 @@
...
+        t.end()
+      })
+    })
+  })
+})
+
+tape('numeric data cache opt creates data cache', function (t) {
+  var feed = create({ cache: { data: 1024 } })
+  var firstLength = Buffer.from('hello').length
+  var secondLength = Buffer.from('world').length
```
mafintosh: Buffer.byteLength('world')

lib/storage.js
andrewosh: Added a test for disabled caching + added checks here

PR #247, opened 2020-02-17
Add failing ifAvailable seek test

andrewosh:
Here's a failing test for ifAvailable seek. If the number of inflight requests is maxed out, the ifAvailable seek will always error with a "Seek not available from peers" error.


mafintosh:
Manually added and fixed

PR #248, opened 2020-02-19
Added update with network partition bug

andrewosh:
Update with two sparsely-syncing peers will hang.


mafintosh:
added and fixed in 8.8.3

PR #250, opened 2020-02-24
Added destroy method for destroying storage

RangerMauve:
This adds a new `destroy` method to the storage which will destroy all its files if possible, and a `destroy` method on the feed which will invoke that instead of `storage.close`


mafintosh:
8.10.0

ZoeDreams:
nice work.

PR #252, opened 2020-03-29
Added peer.on('close')

RangerMauve:
This makes it easier to deal with peers disconnecting in extension messages and in general when you're tracking a peer object directly.


RangerMauve:
Whoops. This doesn't work since Peer objects aren't EventEmtiters

PR #254, opened 2020-04-06
Fix typo (this -> self) in Feed constructor

### LINKED ISSUE #253: Possible typo in Feed constructor
I've just run into an error `this.emit is not a function`, which occured when both `eagerUpdate` and `sparse` are enabled. I'm pretty sure that's due to this line, https://github.com/mafintosh/hypercore/blob/c5cec3badbcfb864dd7ccf6fb38350ff10deb360/index.js#L127 It looks like `this` is pointing to the `loop` function, instead of the `Feed`.

IanS5:
Emit `update-error` from Feed instead of `loop` function on update callback when `eagerUpdate` and `sparse` are enabled.
Fixes #253.

### DIFF
--- a/index.js
+++ b/index.js
@@ -124,7 +124,7 @@ function Feed (createStorage, key, opts) {
 
   if (this.sparse && opts.eagerUpdate) {
     this.update(function loop (err) {
-      if (err) this.emit('update-error', err)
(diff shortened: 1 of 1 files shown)


mafintosh:
Thanks!

PR #255, opened 2020-04-09
Rename .destroy to .destroyData

RangerMauve:
This should help make it less confusing since `.destroy()` is used in place of `.close()` in a lot of places.

Deprecated `.destroy` for now since there's some code using it already. 😅


RangerMauve:
ping @mafintosh @andrewosh

PR #259, opened 2020-05-12
v9

mafintosh:
- [x] Update Noise protocol to latest once @emilbayes lands it.
- [x] Release https://github.com/mafintosh/hypercore-crypto/pull/5
- [x] Release https://github.com/mafintosh/simple-hypercore-protocol/pull/3


mafintosh:
Implements https://github.com/mafintosh/hypercore/issues/260

tinchoz49:
Awesome work!! 

Tell me if I'm wrong but this means that sodium-javascript will be compatible soon right? It's already has support for `crypto_scalarmult` and `chacha20` support seems to be on the way https://github.com/sodium-friends/sodium-javascript/pull/21

mafintosh:
@tinchoz49 indeed!

mafintosh:
Forgot to mention that inspired by a conversation with @noffle yesterday the discovery key hash now includes the wire protocol type. This makes it a lot easier to run multiple versions of hypercore that are incompat as they'll make different swarms

The previous dkey was:

```
blake2b(feedPublicKey, 'hypercore')
```

And the latest one is

```
blake2b(feedPublicKey, 'hypercore 9')
```

We'll only update the `9` part if we ever (unlikely) had to break anything in the wire again.

noffle:
Hey @mafintosh won't this cause problems for hypercore storage modules like [corestore](https://github.com/andrewosh/corestore/) and [multifeed-storage](https://github.com/kappa-db/multifeed-storage), who store hypercores based on each core's discovery key? I don't know corestore that well, but with mf-storage (which isn't yet deployed anywhere afaik) it could result in ALL feeds being duplicated, since it will appear that a remote is sharing a whole new set of feeds.

mafintosh:
@noffle it would and it did shortly until we reverted that part quickly last night! meant to update it here also but the release day stress took over.

noffle:
@mafintosh Sorry, just to clarify: you won't be changing how discovery keys are hashed now?

mafintosh:
@noffle no, we reverted that part due the issues you mentioned.

PR #261, opened 2020-05-13
Add batch support in createReadStream

tinchoz49:
It adds support for batch reading in createReadStream and improve the performance.

It's an initial idea and feedback is welcome.

#### Usage

```javascript
feed.createReadStream({ batch: 100 }) // read batch of 100 messages
```

#### Benchmark

```
> node read-64kb-blocks-linear.js
685658891 bytes/s
10448 blocks/s

> node read-64kb-blocks-linear-batch.js // batch 100 
1080223162 bytes/s
16384 blocks/s
```

#### Todo

- [x] Add test (probably I will use the same tests that we use to test createReadStream)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1308,10 +1308,8 @@ Feed.prototype.createReadStream = function (opts) {
   var live = !!opts.live
   var snapshot = opts.snapshot !== false
   var batch = opts.batch || 1
-  var batchResults = []
   var batchEnd = 0
   var batchLimit = 0
-  var batchIndex = 0
 
   var first = true
   var range = this.download({start: start, end: end, linear: true})
@@ -1332,10 +1330,6 @@ Feed.prototype.createReadStream = function (opts) {
       first = false
     }
 
-    if (batchResults.length > 0 && batchIndex < batchResults.length) {
-      return cb(null, batchResults[batchIndex++])
-    }
-
     if (start === end || (end === -1 && start === self.length)) {
       return cb(null, null)
     }
@@ -1345,7 +1339,6 @@ Feed.prototype.createReadStream = function (opts) {
       return
     }
 
-    batchIndex = 0
     batchEnd = start + batch
     batchLimit = end === Infinity ? self.length : end
     if (batchEnd > batchLimit) {
@@ -1363,8 +1356,11 @@ Feed.prototype.createReadStream = function (opts) {
         return
       }
 
-      batchResults = result
-      cb(null, batchResults[batchIndex++])
+      var lastIdx = result.length - 1
+      for (var i = 0; i < lastIdx; i++) {
+        this.push(result[i])
+      }
+      cb(null, result[lastIdx])
     })
   }
 
--- a/index.js
+++ b/index.js
@@ -1378,13 +1378,13 @@ Feed.prototype.createReadStream = function (opts) {
   }
 
   function setStart (value) {
-    var _start = start
+    var prevStart = start
     start = value
     range.start = start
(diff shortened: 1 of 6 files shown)


tinchoz49:
My concern about this is the download progress. It's not the same to wait for a single message to be download than 100 messages.

Maybe I could check if I have the range of messages downloaded for the batch I choose that option and if is not I use `get`.

tinchoz49:
> My concern about this is related with the download progress. It's not the same to wait for a single message to be download than 100 messages.
Maybe I could check if I have the range of messages downloaded for the batch I choose that option and if is not I use get.

Now I check for the range downloaded and if is not downloaded it use `get`

mafintosh:
@tinchoz49 perhaps we could do a more opportunistic thing where we just check how many messages we have on disk compared to where we are and let's say we have the next 100 messages already downloaded we use the batch api to get them out

CHANGELOG.md
```diff
@@ -9,3 +9,4 @@ See [UPGRADE.md](UPGRADE.md) for notes on breaking changes for downstream develo
 - Ease of use update to signatures, https://github.com/mafintosh/hypercore/issues/260
 - Updates [noise-protocol](https://github.com/emilbayes/noise-protocol) to latest, which uses chacha instead of xchacha and moves avoid from the sodium kx api for better compatability with the rest of the Noise ecosystem.
 - Updates [sodium-native](https://github.com/sodium-friends/sodium-native) from 2 to 3 across the board. 3 uses n-api, so no more builds needed when Node is released.
+- Update the discovery key to hash in the protocol version for better migration
```
mafintosh, line 12: this is from the first commit which i force pushed out. can you remove that commit as well?

index.js
```diff
@@ -1346,6 +1376,16 @@ Feed.prototype.createReadStream = function (opts) {
       read(size, cb)
     })
   }
+
+  function setStart (value) {
+    var _start = start
```
mafintosh, line 1381: _start -> prevStart

test/streams.js
```diff
@@ -3,121 +3,126 @@ var collect = require('stream-collector')
...
+    var feed2 = create()
+
+    feed1.append(['hello', 'world'], function () {
+      var r = feed1.createReadStream({ batch })
+      var w = feed2.createWriteStream()
+
+      r.pipe(w).on('finish', function () {
+        collect(feed2.createReadStream({ batch }), function (err, data) {
+          t.error(err, 'no error')
+          t.same(data, [bufferFrom('hello'), bufferFrom('world')])
```
mafintosh, line 18: Unrelated to this PR this made me realise we need this https://github.com/hypercore-protocol/hypercore/issues/265

mafintosh:
Looks GREAT @tinchoz49 !! Seems to be much faster as well.
I've added a few nits if you wanna take a look.

CHANGELOG.md
tinchoz49: droped it

index.js
tinchoz49: updated

mafintosh:
@tinchoz49 Thanks, I'll land and release this tmw :)

If you wanna post your perf stats here we can include it in the changelog

tinchoz49:
Awesome! thanks for taking the time to review this.

#### Read 100000 Messages 
- createReadStream without batch: `867.438ms`
- createReadStream with batch = 100: `113.650ms`

tinchoz49:
I forgot to add the batch option to the readme

PR #262, opened 2020-05-14
Add upgrade notes for downstream folx

noffle:
Introduces `UPGRADE.md`, a document with details for downstream devs to understand breaking changes and how to integrate these changes into their modules smoothly.


mafintosh:
Thanks! We should add the protocol version thing soon I think. Let’s get an issue started on that

PR #266, opened 2020-05-18
Fix replication benchmarks

fwip:
These were broken since commit a4e5142 corresponding to the v8.0.0 release. Hypercore-protocol version was bumped to 7, which requires the addition of the `initiator` parameter for replication.


mafintosh:
Thank you!!

PR #270, opened 2020-05-25
fix unbound this in createReadStream()

### LINKED ISSUE #269: 'TypeError: this.push is not a function' when setting batch  > 1 in feed.createReadStream()
If you set `opts.batch` to something greater than `1`, this error is triggered. The cause is an unbound `this`: https://github.com/hypercore-protocol/hypercore/blob/master/index.js#L1361

A possible fix:

```
diff --git a/index.js b/index.js
index f1071f2..2132634 100644
--- a/index.js
+++ b/index.js
@@ -1314,7 +1314,8 @@ Feed.prototype.createReadStream = function (opts) {
   var first = true
   var range = this.download({ start: start, end: end, linear: true })
 
-  return from.obj(read).on('end', cleanup).on('close', cleanup)
+  var stream = from.obj(read).on('end', cleanup).on('close', cleanup)
+  return stream
 
   function read (size, cb) {
     if (!self.opened) return open(size, cb)
@@ -1358,7 +1359,7 @@ Feed.prototype.createReadStream = function (opts) {
 
       var lastIdx = result.length - 1
       for (var i = 0; i < lastIdx; i++) {
-        this.push(result[i])
+        stream.push(result[i])
       }
       cb(null, result[lastIdx])
     })

````

jwerle:
fixes #269

### DIFF
--- a/index.js
+++ b/index.js
@@ -1314,7 +1314,8 @@ Feed.prototype.createReadStream = function (opts) {
   var first = true
   var range = this.download({ start: start, end: end, linear: true })
 
-  return from.obj(read).on('end', cleanup).on('close', cleanup)
+  var stream = from.obj(read).on('end', cleanup).on('close', cleanup)
+  return stream
 
   function read (size, cb) {
     if (!self.opened) return open(size, cb)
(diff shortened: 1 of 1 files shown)


mafintosh:
Just noting here as in the issue, that this seems to be a problem for some transpilers, and not an actual bug. Will merge anyway.

jwerle:
@mafintosh thanks!

PR #273, opened 2020-06-08
add maxBlockSize to write stream

mafintosh:


index.js
```diff
@@ -1495,6 +1511,11 @@ Feed.prototype._append = function (batch, cb) {
     var data = this._codec.encode(batch[i])
     var nodes = this._merkle.next(data)
 
+    // the replication stream rejects frames >8MB for DOS defense. Is configurable there, so
```
jwerle, line 1514: Ran into this with `hypercore-multipart`. We're moving files that are 500GB - ~1TB and needed to maximize the block size. Is this documented anywhere else?

PR #274, opened 2020-07-11
Exposing .codecs to turn a given config into a codec.

martinheidegger:
This PR exposes the internal `toCodec` function, adds documentation and a simple test for it.

I noticed in [`crypto-encoder`](https://github.com/ameba23/crypto-encoder/blob/master/index.js#L54) that it contains a poor re-implementations of the `toCodec` function used internally in hypercore. As it is not quite equal to the `codecs` package I thought it would be a good idea to expose it, in order for packages like `crypto-encoder` to integrate more smoothly into the ecosystem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,8 @@ Per default hypercore uses [random-access-file](https://github.com/mafintosh/ran
 
 You can also set valueEncoding to any [abstract-encoding](https://github.com/mafintosh/abstract-encoding) instance.
 
+Using `hypercore.codecs(codec)` you can manually lookup the `abstract-encoding` for a given input.
+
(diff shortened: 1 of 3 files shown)


mafintosh:
Hmm, good question. A little hesitant to expand the API surface as this is mostly only relevant for internal encodings, ie. just a bit nicer to look at raw data files due to this. ndjson and json are fully compatible if you swap them. For example in the crypto case above, they should just use json and not worry about ndjson as they are encrypting anyway.

martinheidegger:
If it would just use `codes` I think it would be fine, but here it applies "hidden-magic" to me. If there were a "hypercore-codecs" package that essentially exposes this magic that would be fine too, though I thought this is less effort and would work.

mafintosh:
we just use compact-encoding now

PR #276, opened 2020-10-28
Using hypercore-streams internally

martinheidegger:
This PR internalizes hypercore-streams and uses streamx consistently.

As `streamx` is not 100% compatible with `readable-streams`, should this be considered a breaking change?

The API did not change but in the tests I needed to add an delay, as `streamx` doesn't immediately create a batch.

~~_It is currently a Draft PR: it requires for several merges:_~~

- [x] https://github.com/mafintosh/merkle-tree-stream/pull/7
- [x] https://github.com/mafintosh/hypercore-streams/pull/2
- [x] https://github.com/mafintosh/hypercore-streams/pull/3


package.json
```diff
@@ -7,30 +7,31 @@
...
     "last-one-wins": "^1.0.4",
     "memory-pager": "^1.0.2",
-    "merkle-tree-stream": "^3.0.3",
+    "merkle-tree-stream": "^4.0.0",
     "nanoguard": "^1.2.0",
     "nanoresource": "^1.3.0",
     "pretty-hash": "^1.0.1",
     "random-access-file": "^2.1.0",
     "sparse-bitfield": "^3.0.0",
+    "streamx": "^2.7.1",
```
mafintosh, line 27: i don't think we use this directly now?

package.json
```diff
@@ -7,30 +7,31 @@
...
     "timeout-refresh": "^1.0.3",
     "uint64be": "^2.0.1",
     "unordered-array-remove": "^1.0.2",
     "unordered-set": "^2.0.0"
   },
   "devDependencies": {
+    "buffer-from": "^1.1.1",
```
mafintosh, line 34: needed?

test/basic.js
```diff
@@ -445,15 +445,17 @@ tape('writes are batched', function (t) {
...
+    ws.write('cd')
+    ws.write('ef')
+    ws.end(function () {
+      t.deepEquals(trackingRam.log.data, [
+        { write: [0, Buffer.from('ab')] },
+        { write: [2, Buffer.from('cdef')] }
+      ])
+      feed.close(function () {
+        t.end()
+      })
```
mafintosh, line 458: would rather change the result of the assert, ie that only one write is performed and then not have the setImmediate.

thats a cool new feature we get from streamx :)

test/streams.js
```diff
@@ -119,7 +119,9 @@ function test (batch = 1) {
         t.fail('should never end')
       })
 
-      feed.append(['c', 'd', 'e'])
+      setImmediate(function () {
+        feed.append(['c', 'd', 'e'])
+      })
```
mafintosh, line 124: why is this needed?

test/basic.js
martinheidegger: Hmm, i think its a good idea to test blocking still (if eventually a block will be formed), is it a good idea to have a test that has two instantly and two after the setImmediate?

package.json
martinheidegger: Will fix, thanks for pointing out!

package.json
martinheidegger: Yes, its needed for the tests. Previously it got installed with one of the removed libraries: https://github.com/hypercore-protocol/hypercore/blob/21e9e666bc182594172c4eb746cfbad6ffdb5ee8/test/streams.js#L4

mafintosh: Nice, makes sense. Yea we should just move it to Buffer.from at some point

test/basic.js
mafintosh: yea that makes sense

test/streams.js
martinheidegger: Same reason as in https://github.com/hypercore-protocol/hypercore/pull/276#discussion_r514252973

test/basic.js
martinheidegger: fixed in https://github.com/hypercore-protocol/hypercore/pull/276/commits/875889d140848a090f8f8bbe199672f1481d77d2

package.json
martinheidegger: fixed in https://github.com/hypercore-protocol/hypercore/pull/276/commits/875889d140848a090f8f8bbe199672f1481d77d2

martinheidegger:
@mafintosh I addressed the open issues in https://github.com/hypercore-protocol/hypercore/pull/276/commits/875889d140848a090f8f8bbe199672f1481d77d2 - can I improve it somehow?

mafintosh:
nope, all good from me

mafintosh:
9.6.0

PR #279, opened 2020-11-06
referencing hypercore-streams docs

martinheidegger:
To have only one source of truth for the hypercore-streams API I figured it is a good idea to just reference the API in order to make sure that users look at the right place for the options.

Companion PR to https://github.com/hypercore-protocol/hypercore-streams/pull/8


mafintosh:

PR #280, opened 2020-11-24
Add byteLength param to upload/download events

pfrazee:
This keeps us equiv to https://github.com/hypercore-protocol/hyperspace/pull/17

There was an undocumented `Peer` parameter being emitted as param 3. That might be a good idea to bring back in the future, but because it wasn't documented and we have no official software that depends on it -- and we already made the decision with remotehypercore in hyperspace -- we're clobbering that param.

### DIFF
--- a/README.md
+++ b/README.md
@@ -512,11 +512,11 @@ Emitted when the feed is ready and all properties have been populated.
 
 Emitted when the feed experiences a critical error.
 
-#### `feed.on('download', index, data)`
+#### `feed.on('download', index, data, byteLength)`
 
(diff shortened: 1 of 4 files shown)


mafintosh:
@pfrazee still relevant?

gmaclennan:
This is useful, however it seems like just syntax sugar, since if I'm not mistaken you can currently do:

```js
on('download', (index, data) => {
  const byteLength = data.length
}
```

I don't think this has any additional cost over passing this as a param?

mafintosh:
fixed in latest

PR #287, opened 2021-02-10
Update: new peer ack event

dpaez:
This PR adds a new event `peer-ack`. It is emitted when `onhave` is called.

Some of the background context for this PR: https://discord.com/channels/709519409932140575/709522073344540804/809152544554483753

In summary, this can be used to know when a peer has something new. This can be useful in the context of clients and seeders.

Related to: https://github.com/RangerMauve/hyperdrive-publisher/pull/10


dpaez:
replaced with #288

PR #291, opened 2021-04-30
feature: onfeedauthenticate to authenticate single feeds per stream

martinheidegger:
Having a stream authentication is super-useful but it may be warranted to authenticate particular feeds to particular peers while others are supposed to work. The proposed API in this PR should work without much overhead, allowing to disable the replication for particular feeds.

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -30,6 +30,44 @@ function replicate (feed, initiator, opts) {
     if (stream.destroyed) return
     if (stream.opened(feed.key)) return
 
+    if (opts.noise !== false && opts.onfeedauthenticate) {
+      if (!stream.remotePublicKey) {
+        feed.ifAvailable.wait()
+        stream.setMaxListeners(0)
+        stream.on('close', onhandshake)
+        stream.on('handshake', onhandshake)
+        return
+      }
+      feedauthenticate()
+      return
+    }
+    replicatePeer()
+  }
+
+  function onhandshake () {
+    feed.ifAvailable.continue()
+    stream.off('close', onhandshake)
+    stream.off('handshake', onhandshake)
+    feedauthenticate()
+  }
+
+  function feedauthenticate () {
+    if (stream.destroyed) return
+    if (stream.opened(feed.key)) return
+    feed.ifAvailable.wait()
+    opts.onfeedauthenticate(feed, stream.remotePublicKey, function (err) {
+      feed.ifAvailable.continue()
+      if (stream.destroyed) return
+      if (stream.opened(feed.key)) return
+      if (err) {
+        stream.close(feed.discoveryKey)
+        return
+      }
+      replicatePeer()
+    })
+  }
+
+  function replicatePeer () {
     if (opts.noise !== false) {
       if (stream.remoteOpened(feed.key) && !stream.remoteVerified(feed.key)) {
         stream.close(feed.discoveryKey)
--- a/README.md
+++ b/README.md
@@ -354,6 +354,9 @@ Options include:
   keyPair: { publicKey, secretKey }, // use this keypair for Noise authentication
   onauthenticate (remotePublicKey, done) // hook that can be used to authenticate the remote peer.
                                          // calling done with an error will disallow the peer from connecting to you.
+  onfeedauthenticate (feed, remotePublicKey, done) // hook similar to onauthenticate but called per feed to replicate over a stream of feeds.
+                                                   // calling done with an error will disallow the peer from syncing this particular feed, but the
+                                                   // stream will stay open.
 }
 ```
 
--- a/test/replicate.js
+++ b/test/replicate.js
@@ -882,6 +882,36 @@ tape('replicate with NOISE disabled', function (t) {
   })
 })
 
+tape('replicate with feed authentication', function (t) {
+  var feed = create()
+  const keysA = Protocol.keyPair()
+  const keysB = Protocol.keyPair()
+  t.plan(5)
+  feed.ready(function (err) {
+    t.error(err)
+    var clone = create(feed.key)
+    const stream = feed.replicate(true, {
+      keyPair: keysA,
+      onfeedauthenticate: function (authFeed, publicKey, cb) {
+        t.equals(authFeed, feed)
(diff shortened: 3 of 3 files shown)


RangerMauve:
This would be amazing to have since we could have really flexible access control restrictions.

lib/replicate.js
```diff
@@ -35,8 +35,27 @@ function replicate (feed, initiator, opts) {
         stream.close(feed.discoveryKey)
         return
       }
+      if (opts.onfeedauthenticate) {
+        if (!stream.remotePublicKey) {
+          stream.once('handshake', function () {
+            onready()
```
mafintosh, line 41: it's not safe to call onready again as it that will mess with the ifAvailable counter

mafintosh: If you call ifAvailable .wait() it might be fine I think

lib/replicate.js
```diff
@@ -35,8 +35,27 @@ function replicate (feed, initiator, opts) {
         stream.close(feed.discoveryKey)
         return
       }
+      if (opts.onfeedauthenticate) {
+        if (!stream.remotePublicKey) {
+          stream.once('handshake', function () {
```
mafintosh, line 40: prob have to setMaxListeners(0) here to avoid leak warnings if we use the event for flow control.

lib/replicate.js
martinheidegger: Thank you for the feedback. I will try to address it, also 😨  I fear that won't be enough as the 'handshake' event may never occur (in case of error, or close) as such i probably need to listen/unlisten to other events as well?

lib/replicate.js
martinheidegger: I was wondering about this: `.setMaxListeners` on stream is _weird_ in this location, shouldn't it be in the `Protocol` constructor of `hypercore-protocol` as the stream may be passed-in from the outside. ([here](https://github.com/hypercore-protocol/hypercore-protocol/blob/c299117fbcbe7100adf442f30768ec4860cacda7/index.js#L358) maybe?) Anyways I will add it.

mafintosh: I normally avoid events for flow control in general because of this, but unsure if you can do this without it. Since this is the one taking on the responsibility of adding that listener, I'd put it here, so it's clear that we aren't just suppressing a leak from the outside.

lib/replicate.js
mafintosh: Yea it's tricky. You need to hook up `on(close)` as well and call the ifavailable hook but ONLY if the happy path hasn't succeded.

martinheidegger:
@mafintosh in the just pushed 9c93de3269d5196db9bb54f5aa4e007c6daee539 I think I managed to take care of closing eventually opened event handlers using ifavailable before/after all possible delays; guarding against closing of stream/parallel replication at any point. Does that work for you?

lib/replicate.js
```diff
@@ -30,6 +30,44 @@ function replicate (feed, initiator, opts) {
...
+    stream.off('close', onhandshake)
+    stream.off('handshake', onhandshake)
+    feedauthenticate()
+  }
+
+  function feedauthenticate () {
+    if (stream.destroyed) return
+    if (stream.opened(feed.key)) return
+    feed.ifAvailable.wait()
+    opts.onfeedauthenticate(feed, stream.remotePublicKey, function (err) {
```
mafintosh, line 58: why is the feed passed actually? is that because this is a corestore feature in practice?

mafintosh:
Yep I think it looks good.

There is an edgecase where the hook runs more than once for the same feed, but I don't think that can be avoided when implemented here. In general it's a more complicated than I prefer, but that's due to the actual stream hook not being exposed. In retrospect it would probably be a lot easier to implement on the replication stream itself as that can have a simple hook here https://github.com/mafintosh/simple-hypercore-protocol/blob/master/index.js#L185 i think, but we can start with this.

Asked a usage question above, cause was a bit confused after rereading it, but other than that it should be fine. Thanks for adding tests.

lib/replicate.js
mafintosh: If we just pass the discoveryKey it's a lot easier to port to the stream layer later on.

martinheidegger: I am passing the feed for an authentication method to have a bit more flexibility in choosing what to authenticate, such as: discoveryKey, publicKey, or size limits etc.

mafintosh: How is that different than doing:

``` js
feed.replicate({
  onfeedauthenticate (remotePublicKey, done) {
    if (feed.....) 
  }
})
```

mafintosh: Ok 👍 It seems a bit weird that we add a "corestore/other-many-core" feature here then, as I think that should be implemented there but we can try it out - ie it might be removed in a future major.

martinheidegger:
> In retrospect it would probably be a lot easier to implement on the replication stream itself as that can have a simple hook here https://github.com/mafintosh/simple-hypercore-protocol/blob/master/index.js#L185 i think, but we can start with this.

I was aware of this possibility, but I don't see the problem in practice: Before the ready hook, `random-access-storage` operations and other delayed checks may also happen more than once, resulting in unintuitive behavior.

Finding the right layer to put things in has been stressful 😅.  The `send` method seems an interesting choice for it _(I wonder what your thought process is)_. I would may have gone with the `onopen` method of `hypercore-protocol` but I am feeling less secure to touch even lower levels of the API 😅

lib/replicate.js
martinheidegger: I hope it won't cause trouble down the road.

mafintosh:
@martinheidegger can you remove the commit from the other PR you made? then this should be good to go.

martinheidegger:
rebased and squashed in https://github.com/hypercore-protocol/hypercore/pull/291/commits/b6de2195dce81033bbf8613bc9d61107d2ab7c57

mafintosh:
Thanks, 9.10.0

mafintosh:
(contains the other pr also)

RangerMauve:
This is super exciting. Thank you both for getting this functionality in. 💜

PR #292, opened 2021-04-30
fixing replicating event to only emit when actually replicating

martinheidegger:
Prior to this PR, the replicating event was emitted whenever `replicate` was called, with this PR its only emitted if the replication is actually happening.


mafintosh:
Makes sense, thanks.

PR #294, opened 2021-07-26
Add custom storage

gmaclennan:
This is a proof-of-concept to see whether Hypercore could support alternative storage to [`random-access-storage`](https://github.com/random-access-storage/random-access-storage).

This PR cleans up the API of the internal [`Storage`](https://github.com/hypercore-protocol/hypercore/blob/master/lib/storage.js) class so that the main hypercore does not reach into the internals of the storage API, then it adds an `opts.storage` option for passing an instance of a custom storage class that implements the same API as the default storage.

NB. This is _not_ the same as `opts.createStorage`, which refers to the instance of [`random-access-storage`](https://github.com/random-access-storage/random-access-storage) used by the `Storage` class. This is for overriding the whole storage with something other than random-access-storage.

### How to implement custom storage?

See [`hypercore-storage-level`](https://github.com/gmaclennan/hypercore-storage-level) as an example storage implemented on a leveldb database.

### Why?

1. **Limited file descriptors**

    Hypercore opens 6 random-access-storage instances, which each leave open a file descriptor. With multiple hypercores open (e.g. for [`multifeed`](https://github.com/kappa-db/multifeed) this can quickly add up to a lot of file descriptors. Android is limited to 1024 FDs per app, and some devices limit to 512, so 86 hypercores would exceed this limit. This is quite possible if each device in a project had 2 hypercores and 50 people are participating.

2. **Server storage**

    For storing Hypercore data online, being tied to random-access-storage requires storage to be on an attached disk. Allowing alternate storage would allow data to be stored in an external database, or object storage such as S3.

3. **Recovering space**

    Currently, using sparse files means that you only use disk space for the hypercore blocks that you have downloaded. However, clearing these blocks does not recover disk space (see https://github.com/mafintosh/fsctl/tree/punchhole for WIP to add native "punchhole" functionality to clear a sparse file and recover disk space). Implementing a custom storage, such as a key-value store, allows disk space to be recovered when blocks are cleared.

### Progress

- [x] **Isolated existing storage module**. Currently hypercore reaches down into the internals of the storage module. Better isolation allows a stable public API for the storage module.
- [x] **Create a proof-of-concept leveldb storage**. For performance reasons it might be better to keep some storage (e.g. bitfield and tree) as random-access storage, but for the initial POC it's using all leveldb.
- [x] **Get tests passing**. Some of the tests reach down into storage internals, so they need to be modified.
    - [x] `test/ack.js`
    - [x] `test/audit.js`
    - [x] `test/basic.js`
    - [x] `test/bitfield.js`
    - [x] `test/cache.js`
    - [x] `test/compat.js`
    - [x] `test/copy.js`
    - [x] `test/default-storage.js`
    - [x] `test/extensions.js`
    - [x] `test/get.js`
    - [x] `test/head.js`
    - [x] `test/replicate.js`
    - [x] `test/seek.js`
    - [x] `test/selections.js`
    - [x] `test/set-uploading-downloading.js`
    - [x] `test/stats.js`
    - [x] `test/streams.js`
    - [x] `test/timeouts.js`
    - [x] `test/tree-index.js`
    - [x] `test/update.js`
    - [x] `test/value-encoding.js`
- [x] Compare benchmarks

| name | before | after | diff |
| ---- | ------ | ----- | ---- |
| write-16kb-blocks | 424,190,256.5 bytes/s<br/>25,848.5 blocks/s | 466,974,804 bytes/s<br/>28,451.5 blocks/s | +10.09% bytes/s<br/>+10.07% blocks/s |
| write-512b-blocks | 86,212,180 bytes/s<br/>168,270 blocks/s | 93,890,196.5 bytes/s<br/>183,036.5 blocks/s | +8.91% bytes/s<br/>+8.78% blocks/s |
| write-64kb-blocks-static | 480,225,178.5 bytes/s<br/>7,321.5 blocks/s | 574,492,573.5 bytes/s<br/>8,759 blocks/s | +19.63% bytes/s<br/>+19.63% blocks/s |
| write-64kb-blocks | 562,612,561.5 bytes/s<br/>8,579.5 blocks/s | 587,978,202 bytes/s<br/>8,961.5 blocks/s | +4.51% bytes/s<br/>+4.45% blocks/s |
| copy-64kb-blocks | 201,904,084.5 bytes/s<br/>3,079 blocks/s | 197,786,239.5 bytes/s<br/>3,017 blocks/s | −2.04% bytes/s<br/>−2.01% blocks/s |
| read-16kb-blocks-proof | 87,754,520.5 bytes/s<br/>5,353.5 blocks/s | 86,934,623.5 bytes/s<br/>5,304 blocks/s | −0.93% bytes/s<br/>−0.92% blocks/s |
| read-16kb-blocks | 752,987,961.5 bytes/s<br/>45,829.5 blocks/s | 746,022,450 bytes/s<br/>45,407 blocks/s | −0.93% bytes/s<br/>−0.92% blocks/s |
| read-512b-blocks | 30,838,337 bytes/s<br/>60,195 blocks/s | 27,954,788 bytes/s<br/>54,569.5 blocks/s | −9.35% bytes/s<br/>−9.35% blocks/s |
| read-64kb-blocks-linear | 742,988,910 bytes/s<br/>11,325 blocks/s | 729,064,189.5 bytes/s<br/>11,116.5 blocks/s | −1.87% bytes/s<br/>−1.84% blocks/s |
| read-64kb-blocks-linear-batch | 764,733,821.5 bytes/s<br/>11,660 blocks/s | 764,505,868 bytes/s<br/>11,652.5 blocks/s | −0.03% bytes/s<br/>−0.06% blocks/s |
| read-64kb-blocks-proof | 327,416,622.5 bytes/s<br/>4,994 blocks/s | 328,411,637.5 bytes/s<br/>5,009 blocks/s | +0.30% bytes/s<br/>+0.30% blocks/s |
| read-64kb-blocks | 2,253,693,382 bytes/s<br/>34,281 blocks/s | 2,344,202,361 bytes/s<br/>35,613.5 blocks/s | +4.02% bytes/s<br/>+3.89% blocks/s |
| replicate-16kb-blocks | 116,052,052 bytes/s<br/>7,079.5 blocks/s | 117,083,592.5 bytes/s<br/>7,141 blocks/s | +0.89% bytes/s<br/>+0.87% blocks/s |
| replicate-64kb-blocks | 236,482,000.5 bytes/s<br/>3,607 blocks/s | 230,973,616.5 bytes/s<br/>3,522 blocks/s | −2.33% bytes/s<br/>−2.36% blocks/s |

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,15 +87,16 @@ Per default hypercore uses [random-access-file](https://github.com/random-access
   secretKey: buffer, // optionally pass the corresponding secret key yourself
   storeSecretKey: true, // if false, will not save the secret key
   storageCacheSize: 65536, // the # of entries to keep in the storage system's LRU cache (false or 0 to disable)
-  onwrite: (index, data, peer, cb) // optional hook called before data is written after being verified
+  onwrite: (index, data, peer, cb), // optional hook called before data is written after being verified
                                    // (remember to call cb() at the end of your handler)
-  stats: true // collect network-related statistics,
+  stats: true, // collect network-related statistics,
   // Optionally use custom cryptography for signatures
   crypto: {
     sign (data, secretKey, cb(err, signature)),
     verify (signature, data, key, cb(err, valid))
-  }
-  noiseKeyPair: { publicKey, secretKey } // set a static key pair to use for Noise authentication when replicating
+  },
+  noiseKeyPair: { publicKey, secretKey }, // set a static key pair to use for Noise authentication when replicating
+  storage: new Storage() // optional storage instance with the same API as [`lib/storage.js`](lib/storage.js). Used for implementing a custom storage such as leveldb. Defaults to an instance of [`lib/storage.js`](lib/storage.js), instantiated with the `createStorage` argument.
 }
 ```
 
--- a/bench/build-comparison-table.js
+++ b/bench/build-comparison-table.js
@@ -1,0 +1,85 @@
+const globby = require('globby')
+const fs = require('fs')
+const fsPromises = fs.promises
+const { format } = require('d3-format')
+
+// Generate a markdown table to compare benchmarks between branches. First write
+// the benchmark results from each branch to a file:
+//
+// ```
+// $ git checkout master
+// $ npm run bench > bench.master.log
+// $ git checkout my-branch
+// $ npm run bench > bench.my-branch.log
+// ```
+//
+// Then generate the table:
+//
+// ```
+// node bench/build-comparison-table.js > bench-comparison-table.md
+// ```
+
+;(async () => {
+  const paths = await globby('branch.*.log')
+
+  const benchmarks = await Promise.all(paths.map(async (filepath) => {
+    const data = await fsPromises.readFile(filepath, 'utf8')
+    const benchmark = parseBenchmarks(data)
+    const [, branch, round] = filepath.match(/^branch\.(.*)\.round-(\d+)/)
+    return {
+      branch: branch === 'master' ? 'before' : 'after',
+      round: Number.parseInt(round),
+      results: benchmark
+    }
+  }))
+  const agg = {}
+
+  for (const { branch, results } of benchmarks) {
+    for (const { name, bytesPerSecond, blocksPerSecond } of results) {
+      agg[name] = agg[name] || {}
+      const result = agg[name][branch] = agg[name][branch] || []
+      result.bytesPerSecond = (result.bytesPerSecond || []).concat(bytesPerSecond)
+      result.blocksPerSecond = (result.blocksPerSecond || []).concat(blocksPerSecond)
+    }
+  }
+
+  const rows = [
+    ['name', 'before', 'after', 'diff'],
+    ['----', '------', '-----', '----']
+  ]
+  for (const [name, { before, after }] of Object.entries(agg)) {
+    const bytesBeforeAvg = avg(before.bytesPerSecond)
+    const bytesAfterAvg = avg(after.bytesPerSecond)
+    const blocksBeforeAvg = avg(before.blocksPerSecond)
+    const blocksAfterAvg = avg(after.blocksPerSecond)
+
+    const bytesDiff = (bytesAfterAvg - bytesBeforeAvg) / bytesBeforeAvg
+    const blocksDiff = (blocksAfterAvg - blocksBeforeAvg) / blocksBeforeAvg
+    rows.push([
+      name,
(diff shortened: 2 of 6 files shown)


pfrazee:
I won't be the one evaluating this, but props for a well-written PR.

gmaclennan:
NB. I completely understand if the team doesn't want to merge this - it exposes more API surface area to maintain, so it might not make sense, and I think there is a slight performance impact. It is super useful though for where we don't have access to the filesystem, e.g. in a browser or in a serverless environment.

martinheidegger:
@gmaclennan `random-access-storage` is not related to the filesystem, it is the base library/system for any kind of implementation and you should be able to create a random-access-leveldb if you wanted to. In the description, I believe you wanted to point at shortcomings of [`random-access-file`](https://github.com/random-access-storage/random-access-file/)?! It would be absolutely possible to manage the amount of file system pointers of random-access-storage by only keeping the (say) last-recent 50 file system pointers at a time open. This could be added to `random-access-file` or even added as a layer on top of `random-access-storage`.

I am thinking of a module called `on-demand-random-access-storage`:

```js
const ondemand = require('on-demand-random-access-storage')
const file = ondemand(require('random-access-file'), {
  maxOpen: 50, // only open up to 50 at a time
  autoClose: 500 // closes idle instances after 500 ms
})
```

This way you could pass any given `random-access-storage` implementation to the `ondemand` constructor and it would open the underlying `random-access-file` implementation only if a `read` or `write` operation happens and close it if idle for too long. Too many requests at a time could use a queue and a LRU to close file handles that have been used "longest time ago".

gmaclennan:
Thanks @martinheidegger, that's certainly one option for solving some of the problems I outline. [`random-access-idb`](https://www.npmjs.com/package/random-access-idb) solves this for indexeddb in the browser by paging data and essentially emulating random access storage on a key-value store.

One of the motivations for experimenting with the approach in this PR is because internally within the storage class of hypercore, it is reading and writing data as if to a key-value store, essentially `storage.getData(sequenceNumber)`. The default storage model essentially does some clever stuff to do that `getData(sequenceNumber)` in a really fast and efficient way over `random-access-storage`, but then emulating random access on a key-value store looses that efficiency.

I think either approach can work. I like that by exposing the storage API of hypercore it allows a lot of flexibility about how data is stored, but it increases the API surface area of hypercore. The changes to hypercore were less work for me that writing `on-demand-random-access-storage` and `random-access-leveldb`, and I think more efficient. I like that it creates options like storing each block of data in a hypercore as a separate file, which is useful when storing large binary data like images. It means that the data can be read without relying on hypercore, since it's just files.

But anyway, putting this out there with an open mind as to whether it makes sense to merge or not.

PR #300, opened 2022-08-16
Add `preupgrade` hook

kasperisager:


mafintosh:
(Some issues with this API in general, so we are punting for now)

PR #302, opened 2022-08-16
feat(core.update()): add option to force update writable cores

Nazeh:
Helps for cautious restoring of Hypercores on new devices, in combination with truncation.

Admittedly for any eagerly replicated and readily seeded core, waiting for a few minutes should be enough to find the latest state, but still.

### DIFF
--- a/index.js
+++ b/index.js
@@ -612,8 +612,7 @@ module.exports = class Hypercore extends EventEmitter {
     if (this.opened === false) await this.opening
     if (this.closing !== null) return false
 
-    // TODO: add an option where a writer can bootstrap it's state from the network also
-    if (this.writable) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -612,8 +612,7 @@ module.exports = class Hypercore extends EventEmitter {
     if (this.opened === false) await this.opening
     if (this.closing !== null) return false
 
-    // TODO: add an option where a writer can bootstrap it's state from the network also
-    if (this.writable) {
+    if (this.writable && !opts?.force === true) {
```
mafintosh, line 615: Can you make this without the ?. syntax? A bit too new for Hypercore.
Otherwise, looks great and love the test :)

Nazeh: Thanks. Done.

PR #309, opened 2022-08-24
Remove t.end() from tests

LuKks:
By the way


LuKks:
@mafintosh Review time!

PR #314, opened 2022-09-06
Switch to `crc-universal`

kasperisager:


mafintosh:
backwards compatible in the end?

kasperisager:
@mafintosh Functionality is the exact same, yes, just with ARM hardware acceleration.

mafintosh:
nice! will test it out tmw

mafintosh:
Just Works ™️

PR #317, opened 2022-09-20
Fix `info.key`, `info.discoveryKey`, and `info.fork`

kasperisager:


kasperisager:

PR #318, opened 2022-09-21
Fix linear downloads due to bitfield indexing bug

kasperisager:
The test case fails when the index is used for linear range downloads. Working on a fix in https://github.com/holepunchto/libquickbit/pull/3.


kasperisager:
Encountered an overflow that was showing up when using the JavaScript fallback as it does bounds checking. Has been fixed in v1.2.3.

PR #321, opened 2022-10-05
Implement remote bitfield indexing

kasperisager:
This PR implements remote bitfield indexing using the new sparse indexing capabilities of Quickbit (https://github.com/holepunchto/libquickbit#sparse-fields). The remote bitfield now maintains a list of segments, each with an index that covers 64 x 4096 KiB pages. Each page is assigned a byte offset within its corresponding segment to translate bit positions to and from the sparse 256 KiB bitfield covered by its segment.

### DIFF
--- a/lib/bitfield.js
+++ b/lib/bitfield.js
@@ -1,6 +1,6 @@
 const BigSparseArray = require('big-sparse-array')
 const b4a = require('b4a')
-const quickbit = require('quickbit-universal')
+const quickbit = require('./compat').quickbit
 
 const BITS_PER_PAGE = 2097152
 const BYTES_PER_PAGE = BITS_PER_PAGE / 8
--- a/lib/bitfield.js
+++ b/lib/bitfield.js
@@ -4,6 +4,7 @@ const quickbit = require('quickbit-universal')
 
 const BITS_PER_PAGE = 2097152
 const BYTES_PER_PAGE = BITS_PER_PAGE / 8
+const WORDS_PER_PAGE = BYTES_PER_PAGE / 4
 
 class BitfieldPage {
   constructor (index, bitfield) {
@@ -51,11 +52,11 @@ module.exports = class Bitfield {
 
     const all = this.resumed
       ? new Uint32Array(buf.buffer, buf.byteOffset, Math.floor(buf.byteLength / 4))
-      : new Uint32Array(BITS_PER_PAGE / 32)
+      : new Uint32Array(WORDS_PER_PAGE)
 
-    for (let i = 0; i < all.length; i += BITS_PER_PAGE / 32) {
-      const bitfield = ensureSize(all.subarray(i, i + (BITS_PER_PAGE / 32)), BITS_PER_PAGE / 32)
-      const page = new BitfieldPage(i / (BITS_PER_PAGE / 32), bitfield)
+    for (let i = 0; i < all.length; i += WORDS_PER_PAGE) {
+      const bitfield = ensureSize(all.subarray(i, i + (WORDS_PER_PAGE)), WORDS_PER_PAGE)
+      const page = new BitfieldPage(i / (WORDS_PER_PAGE), bitfield)
       this._pages.set(page.index, page)
     }
   }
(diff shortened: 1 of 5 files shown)


kasperisager:

lib/bitfield.js
```diff
@@ -48,62 +47,69 @@ module.exports = class Bitfield {
...
       ? new Uint32Array(buf.buffer, buf.byteOffset, Math.floor(buf.byteLength / 4))
-      : new Uint32Array(this._pageSize / 32)
+      : new Uint32Array(BITS_PER_PAGE / 32)
 
-    for (let i = 0; i < all.length; i += this._pageSize / 32) {
-      const bitfield = ensureSize(all.subarray(i, i + (this._pageSize / 32)), this._pageSize / 32)
-      const page = new BitfieldPage(i / (this._pageSize / 32), bitfield)
+    for (let i = 0; i < all.length; i += BITS_PER_PAGE / 32) {
+      const bitfield = ensureSize(all.subarray(i, i + (BITS_PER_PAGE / 32)), BITS_PER_PAGE / 32)
+      const page = new BitfieldPage(i / (BITS_PER_PAGE / 32), bitfield)
```
mafintosh, line 58: mb a UINT32S_PER_PAGE that is just the /32 constant also?

lib/remote-bitfield.js
```diff
@@ -1,63 +1,224 @@
...
+      const m = j + ((i - j) >>> 1)
+
+      if (page.index < this.chunks[m].index) {
+        i = m
+      } else {
+        j = m
+      }
+    }
+
+    this.chunks.splice(i, 0, { field: page.bitfield, offset: page.offset })
```
mafintosh, line 80: since the splice is linear anyway you can just do this like so

```js
chunks.push(x)
for (let i = chunks.length - 2; i >= 0; i--) {
  const prev = chunks[i]
  if (prev <= x) break
  chunks[i] = x
  chunks[i + 1] = prev
} 
```

lib/remote-bitfield.js
```diff
@@ -1,63 +1,224 @@
...
-    const j = index & (this._pageSize - 1)
-    const i = (index - j) / this._pageSize
+    const j = index & (BITS_PER_PAGE - 1)
+    const i = (index - j) / BITS_PER_PAGE
+
+    let p = this._pages.get(i)
 
-    const p = this._pages.get(i) || this._pages.set(i, new Uint32Array(this._pageSize / 32))
+    if (!p && val) {
+      const k = i / PAGES_PER_SEGMENT | 0
```
mafintosh, line 106: I don't think you can do bitwise here due to 32int max (i is unbounded)

lib/remote-bitfield.js
```diff
@@ -1,63 +1,224 @@
...
-    let i = (start - j) / this._pageSize
+    let j = start & (BITS_PER_PAGE - 1)
+    let i = (start - j) / BITS_PER_PAGE
 
     while (length > 0) {
-      const p = this._pages.get(i) || this._pages.set(i, new Uint32Array(this._pageSize / 32))
+      let p = this._pages.get(i)
+
+      if (!p && val) {
+        const k = i / PAGES_PER_SEGMENT | 0
```
mafintosh, line 123: same as above

lib/remote-bitfield.js
kasperisager: Great point!

kasperisager:
@mafintosh Also just realised that the segments should be consulted first during `findFirst()` and `findLast()` so we don't do a linear scan of pages.

lib/remote-bitfield.js
```diff
@@ -57,14 +57,18 @@ class RemoteBitfieldPage {
 class RemoteBitfieldSegment {
   constructor (index) {
     this.index = index
+    this.offset = index * BYTES_PER_SEGMENT
     this.tree = quickbit.Index.from([])
+    this.pages = new Array(64)
```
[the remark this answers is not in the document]
mafintosh: I think this is fine, 64 * 8b, is 512b static overhead, so pretty small.

PR #322, opened 2022-10-09
make replication session close when channel closes

mafintosh:
Also fixes the (false positive) eventemitter warning

### DIFF
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -243,12 +243,14 @@ class BlockTracker {
 }
 
 class Peer {
-  constructor (replicator, protomux, channel) {
+  constructor (replicator, protomux, channel, session) {
     this.core = replicator.core
     this.replicator = replicator
     this.stream = protomux.stream
     this.protomux = protomux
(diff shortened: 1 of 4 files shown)


lib/replicator.js
```diff
@@ -1506,21 +1510,21 @@ module.exports = class Replicator {
...
+  attachTo (protomux, session) {
+    const makePeer = this._makePeer.bind(this, protomux, session)
 
     protomux.pair({ protocol: 'hypercore/alpha', id: this.discoveryKey }, makePeer)
-
     this._ifAvailable++
     protomux.stream.opened.then((opened) => {
       this._ifAvailable--
       if (opened) makePeer()
+      else if (session) session.close().catch(noop)
```
andrewosh, line 1521: Double-checking that this is the right closed condition

mafintosh: Yep, the stream was destroyed before channels could be opened -> kill the replication session

andrewosh (approved):

PR #323, opened 2022-10-10
Update `quickbit-universal`

kasperisager:
The latest patch release makes use of the new `allo`/`allz` functions with SSE 4.1 implementations on Intel, bringing performance on Intel on par with ARM.


kasperisager:

PR #324, opened 2022-10-12
Don't read empty disk pages when loading bitfield

kasperisager:
When loading the bitfield from disk, pass the tree and only load up to its length ceiled to a multiple of 4 KiB.


kasperisager:

mafintosh:
lgtm

PR #325, opened 2022-10-13
Segment local bitfield pages into 4 KiB chunks for improved disk performance

kasperisager:
Pages are back to 32,768 bits (4 KiB), but the underlying buffer is sliced from the associated segment. Flushing a bitfield therefore now only ever writes 4 KiB chunks to disk. Segments contain a dense index and a single contiguous bitfield buffer. The very first segment starts out at 4 KiB and is reallocated and reindexed as pages are added.


kasperisager:

kasperisager:
Which these changes, the remote and local bitfields are _very_ similar, so there might be a future refactor lurking.

PR #328, opened 2022-10-25
Add more replicator tests

LuKks:
Is this behaving as it should?


LuKks:
Before it's merged, maybe it's worth considering to comment out the `t.fail()` on the `ondiscoverykey` callbacks, in case it's somehow fixed in the future so later it's not a breaking change.

At least, the tests are cool as we're adding tests for `createProtocolStream`, and examples on how to use it.

mafintosh:
the ondiscoverkey is working as intended. only fires when a key is received that is not currently paired

PR #332, opened 2022-11-14
Run `test:generate`

kasperisager:


kasperisager:

mafintosh:
good catch

PR #333, opened 2022-11-14
Add `core.batch()` API

kasperisager:


kasperisager:

kasperisager:
This should ideally write blocks to disk rather than keep them in memory and then delay updating the tree until the blocks are committed. Might be for a separate PR though.

lib/batch.js
```diff
@@ -0,0 +1,110 @@
...
+
+    try {
+      if (this._appends.length) await this.session.append(this._appends)
+    } finally {
+      this._clearBatch()
+      this._clearAppends()
+    }
+  }
+
+  async destroy () {
```
mafintosh, line 94: can we name this close for parity with a session?

lib/batch.js
```diff
@@ -0,0 +1,110 @@
...
+
+module.exports = class Batch {
+  constructor (session) {
+    this.session = session
+    this.flushed = false
+
+    this._appends = []
+    this._byteLength = 0
+  }
+
```
mafintosh, line 11: can we add ready () that just forwards to session.ready()

PR #335, opened 2022-11-16
Optionally compute storage estimates in `core.info()`

kasperisager:
Depends on https://github.com/random-access-storage/random-access-memory/pull/15.


kasperisager:

PR #337, opened 2022-11-18
when errors happen check if they were actually conflicts

mafintosh:


test/conflicts.js
```diff
@@ -0,0 +1,36 @@
...
+
+  await unreplicate(streams)
+
+  replicate(c, b, t)
+
+  await t.exception(() => b.get(4))
+
+  // await b.update()
+  // await b.get(2)
+  // await a.truncate(0, 1)
```
kasperisager, line 35: Leftover?

test/conflicts.js
```diff
@@ -0,0 +1,36 @@
...
+    t.is(length, 5, 'conflict at 5 seen by b')
+  })
+
+  await b.get(2)
+
+  await unreplicate(streams)
+
+  replicate(c, b, t)
+
+  await t.exception(() => b.get(4))
```
kasperisager, line 31:
```suggestion
  await t.exception(b.get(4))
```

Should be sufficient given that the value is a promise that'll reject.

kasperisager (changes requested):

test/conflicts.js
mafintosh: ah yes

mafintosh: fixed

test/conflicts.js
mafintosh: fixed

PR #344, opened 2023-01-14
Fix examples/http.js

LuKks:


LuKks:
@mafintosh error seems unrelated

mafintosh:
ya fixed in master

examples/http.js
```diff
@@ -28,14 +28,14 @@ class ByteStream extends streamx.Readable {
     }
 
     if (this.byteOffset > 0) {
-      const [block, byteOffset] = await core.seek(this.byteOffset)
+      const [block, byteOffset] = await this.core.seek(this.byteOffset)
       this.byteOffset = 0
       this.index = block + 1
       this._select(this.index)
-      data = (await core.get(block)).slice(byteOffset)
+      data = (await this.core.get(block)).slice(byteOffset)
```
mafintosh, line 35: can you change this to subarray? better that way

examples/http.js
```diff
@@ -84,7 +84,7 @@ async function start () {
   // hack until we update the replicator
   core.ready = (cb) => cb(null)
 
-  replicator(core, {
+  new Replicator(core, { // eslint-disable-line no-new
```
LuKks, line 87: Replicator needed the `new` otherwise I would have leave it as it was

PR #345, opened 2023-01-15
(draft) Add timeout option for core.get()

LuKks:
It currently has an issue: solo the second test, it passes but the process exits due the promise not being fully handled (I think it's the `await req` on `_cacheOnResolve`)

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,7 +15,7 @@ const BlockEncryption = require('./lib/block-encryption')
 const Info = require('./lib/info')
 const Download = require('./lib/download')
 const { ReadStream, WriteStream } = require('./lib/streams')
-const { BAD_ARGUMENT, SESSION_CLOSED, SESSION_NOT_WRITABLE, SNAPSHOT_NOT_AVAILABLE } = require('./lib/errors')
+const { BAD_ARGUMENT, SESSION_CLOSED, SESSION_NOT_WRITABLE, SNAPSHOT_NOT_AVAILABLE, REQUEST_TIMEOUT } = require('./lib/errors')
 
 const promises = Symbol.for('hypercore.promises')
 const inspect = Symbol.for('nodejs.util.inspect.custom')
@@ -749,6 +749,14 @@ module.exports = class Hypercore extends EventEmitter {
       const req = this.replicator.addBlock(activeRequests, index)
 
       block = this._cacheOnResolve(index, req.promise, this.core.tree.fork)
+
+      if (opts && opts.timeout) {
+        const timeoutId = setTimeout(() => {
+          req.context.detach(req, REQUEST_TIMEOUT())
+        }, opts.timeout)
+
+        req.promise.finally(() => clearTimeout(timeoutId))
+      }
     }
 
     return block
--- a/test/timeouts.js
+++ b/test/timeouts.js
@@ -1,0 +1,38 @@
+const test = require('brittle')
+const { create } = require('./helpers')
+const b4a = require('b4a')
+
+test('get before timeout', async function (t) {
+  t.plan(1)
+
+  const core = await create()
+
+  const req = core.get(0, { timeout: 500 })
+
+  req.then((block) => {
+    t.alike(block, b4a.from('hi'))
+  })
+
+  req.catch((err) => {
+    t.fail(err.message)
+  })
+
+  await core.append('hi')
+})
+
+test('get after timeout', async function (t) {
+  t.plan(1)
+
+  const core = await create()
+
+  const req = core.get(0, { timeout: 500 })
+
+  req.then((block) => {
+    console.log('block', block)
+    t.fail('should not have got block')
+  })
+
+  req.catch((err) => {
+    t.is(err.code, 'REQUEST_TIMEOUT')
+  })
+})
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -52,6 +52,10 @@ module.exports = class HypercoreError extends Error {
     return new HypercoreError(msg, 'REQUEST_CANCELLED', HypercoreError.REQUEST_CANCELLED)
   }
 
+  static REQUEST_TIMEOUT (msg = 'Request timed out') {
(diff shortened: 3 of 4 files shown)


test/timeouts.js
```diff
@@ -0,0 +1,161 @@
...
+
+test('block request gets cancelled before timeout', async function (t) {
+  t.plan(1)
+
+  const core = await create()
+
+  const promise = core.get(0, { timeout: 500 })
+
+  const b = core.replicator._blocks.get(0)
+  b.detach(b.refs[0])
```
LuKks, line 154: Do we remove this test? Due using internals

LuKks:
@mafintosh Ready! Although `r.context` seems to be `this`, so maybe the code can be improved a bit more!

test/timeouts.js
mafintosh: use a session to mimick this instead, ie, close() on a session cancel pending requests

lib/replicator.js
```diff
@@ -54,6 +55,7 @@ class Attachable {
     if (r.rindex < this.refs.length) this.refs[rh.rindex = r.rindex] = rh
     if (r.sindex < r.session.length) r.session[sh.sindex = r.sindex] = sh
 
+    clearTimeout(r.timeout)
```
mafintosh, line 58: if (r.timeout !== null) clearTimeout(r.timeout)
r.timeout = null

lib/replicator.js
```diff
@@ -84,6 +86,15 @@ class Attachable {
...
     }
   }
+
+  setTimeout (r, ms) {
+    if (typeof ms !== 'number') return
+    if (r.timeout !== null) return
+
+    r.timeout = setTimeout(() => {
+      if (r.context) r.context.detach(r, REQUEST_TIMEOUT())
+    }, ms)
```
mafintosh, line 96: r -> req

lib/replicator.js
```diff
@@ -84,6 +86,15 @@ class Attachable {
...
+  }
```
mafintosh, line 97: define a static function

```
function onrequesttimeout(req) {
  if (req.context) ...
}
```

and then update the above to be req.timeout = setTimeout(onrequesttimeout, ms, req)

lib/replicator.js
LuKks: All other methods uses `r`, but I can use `req` on the static function, is that ok?

lib/replicator.js
```diff
@@ -84,6 +87,17 @@ class Attachable {
       this._detach(this.refs[this.refs.length - 1]).reject(err)
     }
   }
+
+  setTimeout (r, ms) {
+    if (r.timeout !== null) {
+      clearTimeout(r.timeout)
+      r.timeout = null
```
LuKks, line 94: Maybe `_destroyTimeout(r)` to reuse it above as well?

mafintosh: Sure! Can also be static as it doesn’t use this

index.js
```diff
@@ -747,6 +747,7 @@ module.exports = class Hypercore extends EventEmitter {
       const activeRequests = (opts && opts.activeRequests) || this.activeRequests
 
       const req = this.replicator.addBlock(activeRequests, index)
+      if (opts && typeof opts.timeout === 'number') req.context.setTimeout(req, opts.timeout)
```
mafintosh, line 750: if opts and opts.timeout now

lib/replicator.js
```diff
@@ -84,6 +87,17 @@ class Attachable {
...
+    }
+
+    if (ms === 0) return
```
mafintosh, line 97: Remove this, that should be in get, see my comment above

test/timeouts.js
```diff
@@ -0,0 +1,147 @@
...
+
+  const core = await create()
+
+  const req = core.get(0, { timeout: 500 })
+
+  req.then((block) => {
+    t.fail('should not have got block: ' + block)
+  }).catch((err) => {
+    t.is(err.code, 'REQUEST_TIMEOUT')
+  })
```
mafintosh, line 32: Use await

test/timeouts.js
```diff
@@ -0,0 +1,147 @@
...
+  try {
+    const block = await b.get(0, { timeout: 1 })
+    t.fail('should not have got block: ' + block)
+  } catch (err) {
+    t.is(err.code, 'REQUEST_TIMEOUT', 'first request failed')
+  }
+
+  setTimeout(() => {
+    a.append('hi')
+  }, 100)
```
mafintosh, line 51: Use setImmediate

test/timeouts.js
```diff
@@ -0,0 +1,147 @@
...
+
+  try {
+    const block = await b1
+    t.fail('should not have got block: ' + block)
+  } catch (err) {
+    t.is(err.code, 'REQUEST_TIMEOUT', 'first request failed')
+  }
+
+  setTimeout(() => {
+    a.append('hi')
```
mafintosh, line 74: setImmediate

test/timeouts.js
```diff
@@ -0,0 +1,147 @@
...
+
+  try {
+    await b1
+  } catch (err) {
+    t.is(err.code, 'REQUEST_TIMEOUT')
+  }
+
+  setTimeout(() => {
+    core.append('hi')
+  }, 100)
```
mafintosh, line 111: setImmediate

test/timeouts.js
```diff
@@ -0,0 +1,147 @@
...
+test('get after timeout with await', async function (t) {
+  t.plan(2)
+
+  const core = await create()
+
+  await core.append('sup')
+  t.alike(await core.get(0, { timeout: 500 }), b4a.from('sup'))
+
+  try {
+    const block = await core.get(1, { timeout: 500 })
```
mafintosh, line 125: Lower it to to 1 the timeout

test/timeouts.js
LuKks: There was different behaviour between using await vs then+catch at one point, but with current PR not relevant anymore

LuKks:
@mafintosh Ready!

PR #346, opened 2023-01-17
allow timeout=0, add catch to finally and check context presence

HDegroote:
I think there were some promise handlers without catch-logic, and that those resulted in the uncaught exception error.

In particular:
-`req.promise.finally` returns a rejected promise if `req.promise` is rejected, so I think it needs to be handled there too (as in: explicitly ignore)
- In the tests, the get(...) path didn't have a catch, so if the promise rejects, this path was also unhandled

Minor things:
- Allowing 0ms timeout (the code was assuming timeout=0 => no timeout, which I don't think is intended)
- I'm not entirely sure, but I think there is a particular flow where req.context is null when setTimeout's method is called, if req.promise resolves in the same cycle as the timeout is triggered. Just in case, I added the check on req.context

### DIFF
--- a/index.js
+++ b/index.js
@@ -750,12 +750,12 @@ module.exports = class Hypercore extends EventEmitter {
 
       block = this._cacheOnResolve(index, req.promise, this.core.tree.fork)
 
-      if (opts && opts.timeout) {
+      if (opts && opts.timeout != null) {
         const timeoutId = setTimeout(() => {
-          req.context.detach(req, REQUEST_TIMEOUT())
+          if (req.context) req.context.detach(req, REQUEST_TIMEOUT())
         }, opts.timeout)
 
-        req.promise.finally(() => clearTimeout(timeoutId))
+        req.promise.finally(() => clearTimeout(timeoutId)).catch(noop)
       }
     }
 
--- a/index.js
+++ b/index.js
@@ -750,7 +750,7 @@ module.exports = class Hypercore extends EventEmitter {
 
       block = this._cacheOnResolve(index, req.promise, this.core.tree.fork)
 
-      if (opts && opts.timeout != null) {
+      if (opts && typeof opts.timeout === 'number') {
         const timeoutId = setTimeout(() => {
           if (req.context) req.context.detach(req, REQUEST_TIMEOUT())
         }, opts.timeout)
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -750,12 +750,12 @@ module.exports = class Hypercore extends EventEmitter {
 
       block = this._cacheOnResolve(index, req.promise, this.core.tree.fork)
 
-      if (opts && opts.timeout) {
+      if (opts && opts.timeout != null) {
```
mafintosh, line 753: typeof opts.timeout === 'number'

test/timeouts.js
```diff
@@ -11,9 +11,7 @@ test('get before timeout', async function (t) {
 
   req.then((block) => {
     t.alike(block, b4a.from('hi'))
-  })
-
-  req.catch((err) => {
+  }).catch((err) => {
```
mafintosh, line 14: use the then(onresolve, onerror) syntax

test/timeouts.js
```diff
@@ -30,9 +28,22 @@ test('get after timeout', async function (t) {
   req.then((block) => {
     console.log('block', block)
     t.fail('should not have got block')
+  }).catch((err) => {
+    t.is(err.code, 'REQUEST_TIMEOUT')
   })
+})
```
mafintosh, line 34: same as above

HDegroote:
Updated. Did notice an issue of clarity in the error traceback: it's not clear at all where it was thrown.

```
const ram = require('random-access-memory')
const Hypercore = require('.')

async function main() {

const core = new Hypercore(ram)
  await core.append('sup')

  console.log(await core.get(0, { timeout: 500 }))
  console.log(await core.get(1, { timeout: 500 }))
}

main()
```

Output:
```
/home/hans/hypercore/index.js:755
          if (req.context) req.context.detach(req, REQUEST_TIMEOUT())
                                                   ^

HypercoreError: REQUEST_TIMEOUT: Request timed out
    at Timeout.<anonymous> (/home/hans/hypercore/index.js:755:52)
    at listOnTimeout (node:internal/timers:564:17)
    at process.processTimers (node:internal/timers:507:7) {
  code: 'REQUEST_TIMEOUT'
}
```

So no reference whatsoever to that line where the error occurs: ` console.log(await core.get(1, { timeout: 500 }))`

Is there an easy way to have a better traceback?

HDegroote:
Everything relevant already adapted in the original PR, so closing

PR #348, opened 2023-01-18
Add constructor timeout, and session inherits timeout

LuKks:


LuKks:
@mafintosh Ready!

test/timeouts.js
```diff
@@ -132,3 +218,7 @@ test('block request gets cancelled before timeout', async function (t) {
 
   await close
 })
+
+function afterTicks (cb) {
+  setImmediate(() => setTimeout(cb, 1))
```
mafintosh, line 223: why not just setImmediate?

LuKks: Ahg, it's a bit difficult to explain, it's like the reason is reversed or something haha

Consider this test:
```js
test('core constructor timeout but disable on get', async function (t) {
  t.plan(1)

  const core = await create({ timeout: 1 })

  setImmediate(() => core.append('hi'))
  t.alike(await core.get(0, { timeout: 0 }), b4a.from('hi'))
})
```

Let's say there was a bug in `{ timeout: 0 }`, so timeout is still 1

So practically the test becomes this:
```js
test('core constructor timeout but disable on get', async function (t) {
  t.plan(1)

  const core = await create({ timeout: 1 })

  setImmediate(() => core.append('hi'))
  t.alike(await core.get(0, { timeout: 1 }), b4a.from('hi'))
})
```

This test will now sometimes work, and sometimes throw

LuKks: So `afterTicks` is only used when we want to ensure that we past the 1ms timeout

LuKks: To summarize, we want this to consistently throw:
```js
setImmediate(() => core.append('hi'))
t.alike(await core.get(0, { timeout: 1 }), b4a.from('hi'))
```

Using this has same effect as previous one (sometimes works, and sometimes throws):
```js
setTimeout(() => core.append('hi'), 2)
t.alike(await core.get(0, { timeout: 1 }), b4a.from('hi'))
```

mafintosh: 👍 Needs a comment in afterTicks tldr’ing that

test/timeouts.js
```diff
@@ -25,6 +53,67 @@ test('get after timeout', async function (t) {
...
+})
+
+test('core constructor timeout but increase on get', async function (t) {
+  t.plan(1)
+
+  const core = await create({ timeout: 1 })
+
+  const get = core.get(0, { timeout: 30000 })
+  setTimeout(() => core.append('hi'), 3)
+  t.alike(await get, b4a.from('hi'))
```
LuKks, line 114: I think it's good now, it first does the `get()` without `await`, later `setTimeout`, and afterwards awaits the first get

test/timeouts.js
```diff
@@ -25,6 +53,67 @@ test('get after timeout', async function (t) {
...
+})
+
+test('core constructor timeout but disable on get', async function (t) {
+  t.plan(1)
+
+  const core = await create({ timeout: 1 })
+
+  const get = core.get(0, { timeout: 0 })
+  setTimeout(() => core.append('hi'), 3)
+  t.alike(await get, b4a.from('hi'))
```
LuKks, line 104: I think it's good now, it first does the `get()` without `await`, later `setTimeout`, and afterwards awaits the first get

LuKks: So for example, if `{ timeout: 0 }` were `1` it should (tm lol) always fail no matter CI slowness, etc

PR #349, opened 2023-01-19
append event emits new length

rafapaezbas:


mafintosh:
Decided, to not land this as I think we should guide people down the same api always.

PR #351, opened 2023-01-20
Add decrypt option to core.get() for disabling

LuKks:


LuKks:
I would still like to reproduce the other case, probably before merging

LuKks:
@mafintosh Ready!

PR #359, opened 2023-02-21
Add seek() options (wait and timeout)

LuKks:
These options will be used for https://github.com/holepunchto/hyperblobs/pull/12

Seek options are needed here:
https://github.com/holepunchto/hyperblobs/blob/2a0b2431ef29af16ad7b59c0564d0eb9619dc542/lib/streams.js#L120


LuKks:
@mafintosh Ready!

PR #360, opened 2023-03-13
Add indexedLength

HDegroote:
Mirrors length, for autobase compat


mafintosh:
10.8.1 since this is "experimental"

PR #363, opened 2023-03-30
Flush Merkle Tree Nodes to Storage in Contiguous Batches

utanapishtim:
I was appending some moderately sized real time media to hypercore in batches, and hit a consistent >1000ms bottleneck flushing merkle nodes to storage per batch. This PR adds logic to flush nodes to disk in batches of contiguous indexes. For my use case it dropped avg latency to <100ms per batch. 

- [x] passes unit tests

### DIFF
--- a/lib/merkle-tree.js
+++ b/lib/merkle-tree.js
@@ -476,10 +476,10 @@ module.exports = class MerkleTree {
       let error = null
       let missing = 1
       let offset = 0
-      let ridx = -1
+      let rindex = -1
 
       const indexes = Array.from(this.flushing.keys()).sort((x, y) => x - y)
-      const rngs = []
+      const ranges = []
 
       // build continuous ranges
       for (const index of indexes) {
@@ -494,18 +494,18 @@ module.exports = class MerkleTree {
         c.uint64.encode(state, node.size)
         c.raw.encode(state, node.hash)
 
-        const rng = rngs[ridx]
-        if (rng && (rng.max + 1) === index) rng.max = index
-        else rngs[++ridx] = { min: index, max: index, offset: offset - 40 }
+        const range = ranges[rindex]
+        if (range && (range.max + 1) === index) range.max = index
+        else ranges[++rindex] = { min: index, max: index, offset: offset - 40 }
       }
 
-      missing += rngs.length
+      missing += ranges.length
 
       // write ranges to disk
-      for (const rng of rngs) {
-        const len = ++rng.max - rng.min
-        const buf = slab.slice(rng.offset, rng.offset + (40 * len))
-        this.storage.write(rng.min * 40, buf, done)
+      for (const range of ranges) {
+        const len = ++range.max - range.min
+        const buf = slab.slice(range.offset, range.offset + (40 * len))
+        this.storage.write(range.min * 40, buf, done)
       }
 
       done(null)
--- a/lib/merkle-tree.js
+++ b/lib/merkle-tree.js
@@ -469,16 +469,22 @@ module.exports = class MerkleTree {
   }
 
   _flushNodes () {
-    // TODO: write neighbors together etc etc
     // TODO: bench loading a full disk page and copy to that instead
     return new Promise((resolve, reject) => {
       const slab = b4a.allocUnsafe(40 * this.flushing.size)
 
       let error = null
-      let missing = this.flushing.size + 1
+      let missing = 1
       let offset = 0
+      let ridx = -1
 
-      for (const node of this.flushing.values()) {
+      const indexes = Array.from(this.flushing.keys()).sort((x, y) => x - y)
+      const rngs = []
+
+      // build continuous ranges
+      for (const index of indexes) {
+        const node = this.flushing.get(index)
+
         const state = {
           start: 0,
           end: 40,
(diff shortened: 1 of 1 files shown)


lib/merkle-tree.js
```diff
@@ -469,16 +469,22 @@ module.exports = class MerkleTree {
...
       const slab = b4a.allocUnsafe(40 * this.flushing.size)
 
       let error = null
-      let missing = this.flushing.size + 1
+      let missing = 1
       let offset = 0
+      let ridx = -1
+
+      const indexes = Array.from(this.flushing.keys()).sort((x, y) => x - y)
+      const rngs = []
```
LuKks, line 482: Probably better to avoid abbreviations, `rngs` -> `ranges`, and below there are also two `rng` -> `range`

Btw really awesome improvement

utanapishtim: @LuKks sounds good, fixed

utanapishtim: Should I change `rindex` to `rangeIndex`?

mafintosh:
Super cool @utanapishtim ! Do you have a simple gist that shows the perf increase?

lib/merkle-tree.js
LuKks: `rindex` is used very simply so it's ok

utanapishtim:
@mafintosh @LuKks sorry for the delay, I haven't been able to replicate the perf increases without access to some of the closed source modules I am using. I tried as best to replicate the env in this branch:

- https://github.com/utanapishtim/hypercore/tree/merkle-tree/optimized-flush-perf

There is a `perf` directory with a README and a set of tests for both browser and native. In the case that prompted me to submit this change, I am running `hypercore` bundled, in the browser, with a closed source `random-access-storage` implementation that has a few 10s ms write latency. My hypothesis is that for slow(er) RAS backends the batching helps to amortize the per-write penalty I am facing. That being said, I couldn't replicate. The perf results I observed in the linked branch are single digit ms improvements and not consistently enough to justify the change.

I am just going to close this PR and try to improve the write latency in the closed source `random-access-storage` implementation.

utanapishtim:
Here is a link to the last set of perf results I got on native, not claiming they are very scientific 😅:

The setup: taking `float32` samples `308700` times per second from my mic and concurrently writing the resulting `16`kb buffers into a hypercore backed by `random-access-memory`. Tests vary in the size of the batch written, the length of time the mic is sampled, and whether the most recently published hypercore or the batched merkle tree version is used.

- with baseline hypercore: https://github.com/utanapishtim/hypercore/blob/merkle-tree/optimized-flush-perf/perf/native/baseline.json
- with merkle tree batched hypercore: https://github.com/utanapishtim/hypercore/blob/merkle-tree/optimized-flush-perf/perf/native/batched.json

As you can see, the results are not consistent enough to justify this change.

LuKks:
In case you ever re-open the PR, I think the code is not using `state` encoding anymore:
https://github.com/utanapishtim/hypercore/blob/4e5ef7724177d69bdea2504ff906cd751acdcf12/lib/merkle-tree.js#L488

`state` should probably be moved below, as otherwise seems wrong to not encode the `node` before writing to storage

utanapishtim:
@LuKks the node is still being encoded on lines `494` and `495`, state is still needed there to make sure we are encoding the node into the correct range of bytes in the slab allocated on line `474`.

utanapishtim:
Ok, I am reopening. I am able to replicate a mocked version of the scenario I am encountering. After further investigation, my scenario is precisely as follows:

- single writes to my storage implementation have a median latency of 22ms
- under contention - eg when many concurrent writes are happening at the same time - this penalty increases linearly with the number of outstanding/pending writes

I've mocked this situation here: https://github.com/utanapishtim/hypercore/blob/merkle-tree/optimized-flush-perf/perf/native/index.js#L14-L28

As you can see, I apply a constant delay on all writes of 22ms and if writes begin to queue the penalty increases linearly with the size of the queue. You can turn [`contentionPenalty`](https://github.com/utanapishtim/hypercore/blob/merkle-tree/optimized-flush-perf/perf/native/index.js#L23) on and off to observe the change.

You can compare the performance between the baseline and batched version of hypercore  with the contention penalty turned on here:

- baseline: https://github.com/utanapishtim/hypercore/blob/merkle-tree/optimized-flush-perf/perf/native/baseline.json
- batched: https://github.com/utanapishtim/hypercore/blob/merkle-tree/optimized-flush-perf/perf/native/batched.json

utanapishtim:
@LuKks and @mafintosh I've recreated a nonmocked equivalent example in the `per/web` directory [here](https://github.com/utanapishtim/hypercore/tree/merkle-tree/optimized-flush-perf/perf/web). The `README` in the `perf` dir has instructions on how to run. The results absolutely demonstrate the order of magnitude speed up in the case of batching. I am going to adjust the PR to see if I can get the best of both worlds, batching only in situations where there are many nodes to flush.

LuKks:
The perf script is a bit complicated, but it's really good that you made a reproducible scenario where you show the maximum performance benefits

`hyperblobs` does batching when writing, just 16 blocks per batch:
https://github.com/holepunchto/hyperblobs/blob/3f6aeefedbb0ecf0c3d32589dc992bf9ee6a0143/lib/streams.js#L84
Will see if we can increase it and/or make it configurable to make more use of your batch optimization at the core level

Meanwhile, you could manually change it to 128 blocks per batch, and by simply comparing `hyperdrive-next` with vs without your `hypercore` branch. For example, it would be cool to compare how much it takes in simple milliseconds writing several 100mb files (cool to have min/max as well)

This will end up resulting in several improvements for writing speed, really awesome!

utanapishtim:
@LuKks @mafintosh everything has been simplified [here](https://github.com/utanapishtim/hypercore/tree/merkle-tree/optimized-flush-perf/perf), scripts are very short and still demonstrative of perf increase. You can tune the parameters how you want to observe changes. Included Hyperblobs perf test as well.

utanapishtim:
@mafintosh @LuKks anything I can do to move this along? Maintaining a fork right now so I can ship this.

PR #364, opened 2023-04-08
enables firewalling replication on a per-peer basis

utanapishtim:
This pull request captures the work required to firewall peers on replication, effectively re-implements https://github.com/holepunchto/hypercore/pull/291 for v10.

It works by:

1. waiting for the noiseStream to connect
2. authenticating the peer
3. attaching the mux and session to the replicator

This allows for finer-grained firewalling of individual hypercores, it mirrors the [hyperdht](https://github.com/holepunchto/hyperdht#const-server--nodecreateserveroptions-onconnection) firewall api.

### open questions
- [x] Does mux need to be corked before and uncorked after firewalling? No.
- [ ] Is this the right api? 

### open work
- [ ] update the README if the api looks right
- [ ] update Corestore to support a firewall option

### DIFF
--- a/index.js
+++ b/index.js
@@ -75,6 +75,7 @@ module.exports = class Hypercore extends EventEmitter {
     this.wait = opts.wait !== false
     this.timeout = opts.timeout || 0
     this.firewall = opts.firewall || null
+    this.onfirewalled = opts.onfirewalled || noop
 
     this.closing = null
     this.opening = this._openSession(key, storage, opts)
@@ -431,7 +432,8 @@ module.exports = class Hypercore extends EventEmitter {
   }
 
   replicate (isInitiator, opts = {}) {
-    opts = { firewall: this.firewall, ...opts }
+    const { firewall, onfirewalled } = this
+    opts = { firewall, onfirewalled, ...opts }
     // Only limitation here is that ondiscoverykey doesn't work atm when passing a muxer directly,
     // because it doesn't really make a lot of sense.
     if (Protomux.isProtomux(isInitiator)) return this._attachToMuxer(isInitiator, { session: true, ...opts })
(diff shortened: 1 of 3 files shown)


utanapishtim:
I made the method async to support - for example - checking whether the remotePublicKey is in a Hyperbee. I think the method over in `hyperdht` should be changed to be async as well. I don't know what you all think about that.

index.js
```diff
@@ -430,35 +431,44 @@ module.exports = class Hypercore extends EventEmitter {
...
     const session = useSession ? this.session() : null
-    this.replicator.attachTo(mux, session)
+
+    const maybeAttach = (allow) => (allow) ? this.replicator.attachTo(mux, session) : mux.destroy()
+
+    if (!opts.firewall) return maybeAttach(true)
+
+    mux.stream.noiseStream.opened.then(() => {
+      const { remotePublicKey, handshake } = mux.stream.noiseStream
+      opts.firewall(remotePublicKey, handshake, this).then(maybeAttach, mux.destroy.bind(mux))
```
utanapishtim, line 470: - [ ] I need to make sure destroying mux is the right thing to do here, it may just be that the session needs to be closed

index.js
```diff
@@ -430,35 +431,47 @@ module.exports = class Hypercore extends EventEmitter {
...
 
-  _attachToMuxerOpened (mux, useSession) {
+  _attachToMuxerOpened (mux, useSession, opts = {}) {
     // If the user wants to, we can make this replication run in a session
     // that way the core wont close "under them" during replication
-    const session = useSession ? this.session() : null
-    this.replicator.attachTo(mux, session)
+
+    const maybeAttach = (allow) => {
+      if (!allow) return mux.destroy()
```
utanapishtim, line 464: @mafintosh I don't think this is correct. Suppose someone passes `replicate` a `Protomux` instance, should we really destroy if replication is firewalled? What if they are running other protocols on the mux instance? It would be nice to signal firewalling a bit more responsibly. If we noop, the caller has no indication that replication has been firewalled. We can't throw because the chain of methods is sync and throwing is ugly, no need to panic on getting firewalled. We could thread `protocolStream` through from `replicate` and emit a `firewalled` event...that's the best I can come up with right now. Some guidance would be appreciated.

utanapishtim:
I added an `onfirewalled` hook that gets passed the firewalled mux instance and the core it was prevented from accessing.

index.js
```diff
@@ -430,35 +432,48 @@ module.exports = class Hypercore extends EventEmitter {
...
+      if (deny) return opts.onfirewalled(mux, this)
+      const session = useSession ? this.session() : null
+      this.replicator.attachTo(mux, session)
+    }
+
+    if (!opts.firewall) return maybeAttach(false)
+
+    mux.stream.noiseStream.opened.then(() => {
+      const { remotePublicKey } = mux.stream.noiseStream
+      opts.firewall(remotePublicKey, this).then(maybeAttach, mux.destroy.bind(mux))
```
utanapishtim, line 475: If we are unable to successfully return a firewall verdict, it feels right to panic on the mux and destroy.

PR #369, opened 2023-05-10
Fix range clamping when downloading blocks

kasperisager:


kasperisager:
@mafintosh This is ready for review. The new test cases are quite slow, so I can mark them as skipped if preferred.

lib/replicator.js
```diff
@@ -1673,17 +1673,13 @@ function clampRange (core, r) {
...
-        r.start = start
-        r.end = end
-      }
+    if (r.end === -1) r.start = start === -1 ? core.tree.length : start
+    else if (start === -1) r.start = r.end
+    else {
+      const end = core.bitfield.lastUnset(r.end)
+
+      r.start = start
+      r.end = end
```
kasperisager, line 1682: This seems a little more conservative than previously, but would not move `end` if `end - 1` is a set bit.

PR #371, opened 2023-05-10
Do linear clamp for range downloads

kasperisager:
Stop-gap measure until I can figure out what's up with the bitfield index.


kasperisager:
Closed in favor of #372.

PR #372, opened 2023-05-11
Fix range clamping on range downloads, take 2

kasperisager:
Closes #371.


lib/replicator.js
```diff
@@ -1674,12 +1674,13 @@ function clampRange (core, r) {
...
+    else if (start === -1 || start >= r.end) r.start = r.end
     else {
-      const end = core.bitfield.lastUnset(r.end)
-
       r.start = start
-      r.end = end
+
+      const end = core.bitfield.lastUnset(r.end - 1)
+
+      if (end !== -1) r.end = end + 1
```
kasperisager, line 1683: This now correctly returns `-1` when no last unset bit is found so we need to guard against that.

PR #373, opened 2023-05-11
Add diff option to core.clear() to return bytes cleared

LuKks:


test/clear.js
```diff
@@ -77,3 +79,19 @@ test('incorrect clear', async function (t) {
...
 })
+
+test('clear with diff option', async function (t) {
+  const storage = createTmpDir(t)
+
+  const core = new Hypercore(storage)
+  const block = await core.append(b4a.alloc(32 * 1024).fill('a'))
+
+  console.log(block)
+  await core.clear(1)
```
LuKks, line 90: Just temporal debugging, it fails with this:

`{ length: 1, byteLength: 32768 }`

```
/self/hypercore/node_modules/fs-native-extensions/index.js:5
  if (errno < 0) this.reject(toError(errno))
                             ^

Error: EINVAL: invalid argument
    at Object.onwork (/self/hypercore/node_modules/fs-native-extensions/index.js:5:30) {
  errno: -22,
  code: 'EINVAL'
}
```

index.js
```diff
@@ -737,9 +737,16 @@ module.exports = class Hypercore extends EventEmitter {
...
+    if (start >= end) return 0
+
+    let bytes = 0
+    if (opts && opts.diff) bytes = (await this.info({ storage: true })).blocks
 
     await this.core.clear(start, end)
+
+    if (opts && opts.diff) bytes -= (await this.info({ storage: true })).blocks
+
+    return bytes
```
mafintosh, line 749: should return { blocks: bytes } so it can tell for each file in the future

mafintosh: can also be done a lot faster, by exposing a static helper from Info to get the bytesUsed and just doing it with the blocks file

LuKks:
Should we return this `length - offset`? Which seems to be the actual total bytes cleared?
https://github.com/holepunchto/hypercore/blob/54d291b9cbc9a103c7d89b50485d82fa9dc5b751/lib/core.js#L274

Maybe `core.clear()` can return `{ bytes, blocks }`

Where `bytes` is the actual exact amount of bytes cleared, but then `blocks` and potentially others are based on the storage info

index.js
```diff
@@ -737,9 +737,21 @@ module.exports = class Hypercore extends EventEmitter {
       end = start + 1
     }
 
-    if (start >= end) return
+    const cleared = (opts && opts.diff) ? { blocks: 0 } : null
+
+    if (start >= end) return cleared
```
LuKks, line 742: Can we safely remove this check?

index.js
```diff
@@ -737,9 +737,21 @@ module.exports = class Hypercore extends EventEmitter {
...
+
+    const before = cleared ? await Info.bytesUsed(this.core.blocks.storage) : null
 
     await this.core.clear(start, end)
+
+    const after = cleared ? await Info.bytesUsed(this.core.blocks.storage) : null
+
+    if (cleared) {
+      cleared.blocks = before - after
```
mafintosh, line 751: lets Math.max this with 0, as the disk is not lockable

LuKks: There is no local lock that we could use? So within the same program works as expected

mafintosh: good point, we should consider moving this into the locked section actually. Still needs the guard, as we dont own the disk.

PR #375, opened 2023-05-12
Support multi key encryption

kasperisager:
This PR introduces a new `encryption` option for customising the logic for how encryption keys and nonces are constructed. The option is an object containing two hooks, `id()` and `key()`, as seen in the new test case:

https://github.com/holepunchto/hypercore/blob/147e6e16353a48a2cd2abb3f0e9c7ab56221fb6c/test/encryption.js#L199-L211

The `id()` hook must encode into `id` 8 bytes of data that when combined with the block sequence number, `index`, uniquely identifies the block. This identifier is then used as a nonce for encrypting and decrypting the block at `index` using the key returned by the `key()` hook.

The default encryption scheme still relies on just a single encryption key, but is now implemented using the new hooks:

https://github.com/holepunchto/hypercore/blob/147e6e16353a48a2cd2abb3f0e9c7ab56221fb6c/lib/block-encryption.js#L61-L94


kasperisager:
This is no longer needed.

PR #376, opened 2023-05-15
Add core.createByteStream([options])

LuKks:


lib/streams.js
```diff
@@ -54,3 +54,62 @@ class WriteStream extends Writable {
 }
 
 exports.WriteStream = WriteStream
+
+class ByteStream extends Readable {
+  constructor (core, byteOffset, byteLength) {
```
mafintosh, line 59: make the byteOffset and byteLength options

package.json
```diff
@@ -57,6 +57,7 @@
     "hyperswarm": "^4.3.6",
     "random-access-memory": "^6.1.0",
     "random-access-memory-overlay": "^3.0.0",
+    "range-parser": "^1.2.1",
```
LuKks, line 60: Dev dependency, maybe I should leave it there as it's used in the examples folder and it's unexpected to run the file to result in module not found

LuKks: The idea was to inline it but there is no much gain in doing so, unless we make a static helper that simplifies handling the Response object, setting the headers in relation to the ranges, etc but that sounds like a new tiny lib rather that inlining it in Hypercore

lib/streams.js
```diff
@@ -54,3 +54,77 @@ class WriteStream extends Writable {
...
+  }
+
+  _open (cb) {
+    this._openp().then(cb, cb)
+  }
+
+  async _openp () {
+    if (this.byteLength === -1) {
+      await this.core.update()
+      this.byteLength = this.core.byteLength
```
mafintosh, line 78: - byteOffset no? (max'ed with 0)

lib/streams.js
```diff
@@ -54,3 +54,77 @@ class WriteStream extends Writable {
...
+      const [block, byteOffset] = await this.core.seek(this.byteOffset)
+
+      this.byteOffset = 0
+      this.index = block
+
+      relativeOffset = byteOffset
+    }
+
+    this._prefetch(this.index + 1)
+    data = await this.core.get(this.index++)
```
mafintosh, line 102: try catch this and return cb(err)

lib/streams.js
mafintosh: i'd just call this open like below and do the same tricks

package.json
mafintosh: sure, dont inline it

lib/streams.js
LuKks: No longer necessary to try-catch it due this?
https://github.com/holepunchto/hypercore/blob/e8f6a8f5dfb4d87e874605801183f20c950dc745/lib/streams.js#L76

lib/streams.js
LuKks: `-1` means max byte length, so unsure if I should apply the minus offset

LuKks: Will add tests, that could show any obvious bug

LuKks: You're right, I thought it was like `byteEnd`

lib/streams.js
```diff
@@ -54,3 +54,79 @@ class WriteStream extends Writable {
...
+      this.push(null)
+      return
+    }
+
+    let relativeOffset = 0
+
+    if (this.byteOffset > 0) {
+      const [block, byteOffset] = await this.core.seek(this.byteOffset)
+
+      this.byteOffset = 0
```
mafintosh, line 99: dont you think this should be _byteOffset since we set it to 0?

lib/streams.js
```diff
@@ -54,3 +54,79 @@ class WriteStream extends Writable {
...
+
+    if (data.length >= this.byteLength) {
+      data = data.subarray(0, this.byteLength)
+      this.push(data)
+      this.push(null)
+    } else {
+      this.push(data)
+    }
+
+    this.byteLength -= data.length
```
mafintosh, line 118: should also be _byteLength I think

lib/streams.js
```diff
@@ -54,3 +54,79 @@ class WriteStream extends Writable {
...
+  constructor (core, opts = {}) {
+    super()
+
+    this.core = core
+    this.byteOffset = opts.byteOffset || 0
+    this.byteLength = typeof opts.byteLength === 'number' ? opts.byteLength : -1
+    this.prefetch = typeof opts.prefetch === 'number' ? opts.prefetch : 32
+
+    this.index = 0
+    this.range = null
```
mafintosh, line 68: all of these -core, should be _ props i think

lib/streams.js
LuKks: True, I added a boolean to avoid mutating it but still `byteLength` needs to be private so not cool to just have one prop

lib/streams.js
LuKks: Yep

lib/streams.js
```diff
@@ -54,3 +54,78 @@ class WriteStream extends Writable {
...
+    }
+
+    this._prefetch(this._index + 1)
+    data = await this._core.get(this._index++)
+
+    if (relativeOffset > 0) data = data.subarray(relativeOffset)
+
+    const final = data.length >= this._byteLength
+    if (final) data = data.subarray(0, this._byteLength)
+    this._byteLength -= data.length
```
mafintosh, line 114: if (data.byteLength > this._byteLength) data = data.subarray(0, this._byteLength)
this._byteLength -= data.byteLength

...


if (this._byteLength === 0) this.push(0)

lib/streams.js
```diff
@@ -54,3 +54,78 @@ class WriteStream extends Writable {
...
+  constructor (core, opts = {}) {
+    super()
+
+    this._core = core
+    this._index = 0
+    this._range = null
+
+    this._byteOffset = opts.byteOffset || 0
+    this._byteLength = typeof opts.byteLength === 'number' ? opts.byteLength : -1
+    this.prefetch = typeof opts.prefetch === 'number' ? opts.prefetch : 32
```
mafintosh, line 68: should be _ also, rename your func to be _updatePrefetch or something

lib/streams.js
LuKks: Can't believe that is equivalent lol

mafintosh:
excellent!

PR #378, opened 2023-05-19
Bugfix non-sparse byteLength

HDegroote:
`this.contiguousByteLength` always returns 0. Removing the special-case on non-spareness fixes it


mafintosh:
Ah configByteLength is simply not implemented atm :)

PR #379, opened 2023-05-19
Add purge

HDegroote:
Still needs to delete the dir itself

### DIFF
--- a/index.js
+++ b/index.js
@@ -742,6 +742,20 @@ module.exports = class Hypercore extends EventEmitter {
     await this.core.clear(start, end)
   }
 
+  async purge () {
+    await this.close()
+    // TODO: unlink dir itself
+
+    const proms = ['oplog', 'tree', 'bitfield', 'blocks'].map(
+      filename => new Promise((resolve, reject) => {
+        const cb = (err, val) => err ? reject(err) : resolve(val)
+        return this.core[filename].storage.unlink(cb)
+      })
+    )
+
+    await Promise.all(proms)
+  }
+
   async _get (index, opts) {
     let block
 
--- a/index.js
+++ b/index.js
@@ -750,7 +750,7 @@ module.exports = class Hypercore extends EventEmitter {
   }
 
   async purge () {
-    await this._closeAllSessions()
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -742,6 +749,11 @@ module.exports = class Hypercore extends EventEmitter {
     await this.core.clear(start, end)
   }
 
+  async purge () {
+    await this._closeAllSessions()
```
mafintosh, line 753: pass null here so its explicit

lib/core.js
```diff
@@ -280,6 +280,33 @@ module.exports = class Core {
...
+        const cb = (err, val) => err ? reject(err) : resolve(val)
+        this.bitfield.storage.unlink(cb)
+      }),
+      new Promise((resolve, reject) => {
+        const cb = (err, val) => err ? reject(err) : resolve(val)
+        this.blocks.storage.unlink(cb)
+      })
+    ]
+
+    await Promise.all(proms)
```
mafintosh, line 307: just do

```js
return new Promise((resolve, reject) => {
  let missing = 4
  let error = null

  this.oplog.storage.unlink(done)
  this.tree.storage.unlink(done)
  this.bitfield.storage.unlink(done)
  this.blocks.storage.unlink(done)

  function done (err) {
    if (err) error = err
    if (--missing) return
    if (error) reject(err)
    else resolve()
  }
})
```

HDegroote: Nice, fixed

test/purge.js
```diff
@@ -0,0 +1,45 @@
...
+  t.is(core.closed, true)
+  t.is(fs.existsSync(oplogLoc), false)
+  t.is(fs.existsSync(treeLoc), false)
+  t.is(fs.existsSync(bitfieldLoc), false)
+  t.is(fs.existsSync(dataLoc), false)
+  t.is(fs.readdirSync(dir).length, 0) // Nothing remains
+})
+
+test('purge closes all sessions', async function (t) {
+  const dir = createTmpDir()
```
mafintosh, line 35: pass the teardown to the tmp helper in all tests no? so the test folders are gc'ed

PR #382, opened 2023-05-24
Use RAM in test: clear blocks with diff option

LuKks:
I thought it wasn't possible for it to return cleared bytes using RAM but seems possible by using a bigger buffer

### DIFF
--- a/test/clear.js
+++ b/test/clear.js
@@ -3,7 +3,7 @@ const b4a = require('b4a')
 const RAM = require('random-access-memory')
 const { create, replicate, eventFlush } = require('./helpers')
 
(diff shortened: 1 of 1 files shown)


LuKks:
Ah I see here:
https://github.com/random-access-storage/random-access-memory/blob/52db73d55d3f4ed011a2b17abbe56fb7ac68be01/index.js#LL5C7-L5C24

mafintosh:
cleaner to make it with an explicit page size imo

mafintosh:
Just for this test

LuKks:
@mafintosh Ready!

PR #384, opened 2023-05-28
Add Apache License and Notice

rafapaezbas:


telamon:
Missing Description Provided:
This pr transfers ownership of hypercore* from the individual to the corporation.
It also modifies the rules for community contribution, requiring contributors to explicitly relinquish all rights to the corporation.
It doesn't mean much right now, and probably never will.
But if it does, now is a good time as any

mafintosh:
(we wanna migrate to apache-2.0 everywhere, but we just opened issues everywhere in one go, hc requires some work before)

PR #386, opened 2023-05-29
Add docs for core.session([options])

LuKks:


mafintosh:
@LuKks looks good. should mention that you need to close any session you make

PR #388, opened 2023-06-04
Adjust hidden _preready for Corestore

LuKks:


mafintosh:
not needed anymore :)

PR #389, opened 2023-06-06
support experimental unlocked mode

mafintosh:


telamon:
This sounds interesting. Would unlock mode enable development of multiple machines with same secret / WiFi bandwidth-lanes allocation style?

mafintosh:
@telamon its all local on the same core

PR #391, opened 2023-06-12
Clear test where it could not load node

LuKks:


mafintosh:
fixed (and added) in #420

PR #393, opened 2023-06-15
Make Hypercore errors a public export

chm-diederichs:
This PR allows users to import hypercore errors as `require('hypercore/errors')`.

This is useful to throw from dependent modules such as Hyperbee and Autobase.


lib/errors.js
```diff
@@ -44,6 +44,10 @@ module.exports = class HypercoreError extends Error {
     return new HypercoreError(msg, 'INVALID_PROOF', HypercoreError.INVALID_PROOF)
   }
 
+  static BLOCK_NOT_AVAILABLE (msg = 'Snapshot is not available') {
```
mafintosh, line 47: lets fix the message

PR #395, opened 2023-06-21
Max inflight based on RTT

LuKks:


lib/replicator.js
```diff
@@ -564,6 +596,9 @@ class Peer {
       if (req.peer !== this) return
       this.inflight--
       this.replicator._inflight.remove(req.id)
+
+      const latency = Date.now() - req.$started
+      this.latencyInflight = this.latencyInflight ? ((this.latencyInflight + latency) / 2) : latency
```
LuKks, line 584: Those things and the `req.$started` is for debugging only, will remove before merging or now if desired

mafintosh:
10.17.0

PR #397, opened 2023-06-22
Add receiver queue for requests

LuKks:


lib/replicator.js
```diff
@@ -504,6 +507,28 @@ class Peer {
   }
 
   async onrequest (msg) {
+    const shouldDrain = this.receiverQueue.length === 0
+
+    this.receiverQueue.push(msg.id, msg)
+
+    if (shouldDrain) await this._drainRequests()
```
mafintosh, line 514: should make this if (not busy) await drain()

lib/receiver-queue.js
```diff
@@ -0,0 +1,44 @@
...
+    this.requests = new Map()
+
+    this.busy = false
+  }
+
+  get length () {
+    return this.queue.length
+  }
+
+  push (id, req) {
```
mafintosh, line 13: req has an id attached already

mafintosh: so just push(req)

lib/receiver-queue.js
```diff
@@ -0,0 +1,44 @@
...
+    if (this.requests.has(id)) throw new Error('Request already exists: ' + id)
```
mafintosh, line 14: just let last one "win"

lib/receiver-queue.js
```diff
@@ -0,0 +1,42 @@
...
+    this.queue.push(req)
+    if (req.id !== 0) this.requests.set(req.id, req)
+  }
+
+  shift () {
+    while (this.queue.length > 0) {
+      const req = this.queue.shift()
+      if (req.id !== 0) this.requests.delete(req.id)
+
+      if (req.block || req.hash || req.seek || req.upgrade) return req
```
LuKks, line 23: Just in case, if id is zero do we want to always return here?

Or zero id requests would always have one of those? Then it's fine as it is

LuKks:
Benchmark here: https://github.com/holepunchto/hypercore/pull/398

With no queue: ~2000ms
With recv queue: ~2850ms

LuKks:
With parallelism of two: ~2200ms

lib/replicator.js
```diff
@@ -73,6 +69,12 @@ class Attachable {
   }
 
   _cancel (r, err) {
+    if (r.context.inflight) {
+      for (const req of r.context.inflight) {
+        req.peer.wireCancel.send({ request: req.id })
```
LuKks, line 74: No test seems to reach this wire cancel, tried to reproduce it but nothing atm

lib/receiver-queue.js
mafintosh: ya this is ok for now i think

lib/replicator.js
```diff
@@ -42,8 +43,8 @@ class Attachable {
   detach (r, err = null) {
     if (r.context !== this) return false
 
-    this._detach(r)
     this._cancel(r, err)
+    this._detach(r)
```
mafintosh, line 47: i think this can stay where it was given our convo

lib/replicator.js
```diff
@@ -510,6 +514,42 @@ class Peer {
...
+    }
+  }
+
+  async _handleRequests () {
+    if (this.receiverBusy) return
+    this.receiverBusy = true
+
+    while (this.receiverQueue.length > 0) {
+      if (!this.remoteOpened) break
+      if (!this.protomux.drained) break
```
mafintosh, line 541: put these two in the while condition?

lib/receiver-queue.js
```diff
@@ -0,0 +1,42 @@
+const FIFO = require('fast-fifo')
+
+module.exports = class ReceiverQueue {
+  constructor () {
+    this.queue = new FIFO()
+    this.requests = new Map()
+  }
+
+  get length () {
+    return this.queue.length
```
mafintosh, line 10: mb maintain your own length so you can dec on internal gc

lib/receiver-queue.js
```diff
@@ -0,0 +1,41 @@
...
+
+    const req = this.requests.get(id)
+    if (!req) return
+
+    req.block = null
+    req.hash = null
+    req.seek = null
+    req.upgrade = null
+
+    this.requests.delete(id)
```
mafintosh, line 39: decrement length here

lib/receiver-queue.js
```diff
@@ -0,0 +1,41 @@
...
+    this.queue.push(req)
+    if (req.id !== 0) this.requests.set(req.id, req)
+    this.length++
+  }
+
+  shift () {
+    while (this.queue.length > 0) {
+      const req = this.queue.shift()
+      if (req.id !== 0) this.requests.delete(req.id)
+      this.length--
```
mafintosh, line 20: only decrement length in the next if

lib/replicator.js
```diff
@@ -111,7 +106,13 @@ class BlockRequest extends Attachable {
   }
 
   _unref () {
-    if (this.inflight.length > 0) return
+    if (this.inflight.length > 0) {
+      for (const req of this.inflight) {
+        req.peer._cancelRequest(req.id)
+      }
+      return
```
mafintosh, line 113: i ... think ... you wanna remove this return, otherwise no one removes it from the tracker

lib/replicator.js
```diff
@@ -510,6 +514,40 @@ class Peer {
   }
 
   async onrequest (msg) {
+    if (!this.protomux.drained) {
+      this.receiverQueue.push(msg)
+      console.log('Pushed to the queue', this.receiverQueue.length)
```
mafintosh, line 519: log

lib/replicator.js
```diff
@@ -510,6 +514,40 @@ class Peer {
   }
 
   async onrequest (msg) {
+    if (!this.protomux.drained) {
```
mafintosh, line 517: if !drained or receiverQueue.length

lib/replicator.js
```diff
@@ -510,6 +514,40 @@ class Peer {
...
+    }
+  }
+
+  async _handleRequests () {
+    if (this.receiverBusy) return
+    this.receiverBusy = true
+
+    while (this.remoteOpened && this.protomux.drained && this.receiverQueue.length > 0) {
+      const msg = this.receiverQueue.shift()
+      if (!msg) continue
```
mafintosh, line 542: no need for this if with the suggested length update above

lib/receiver-queue.js
mafintosh: and if length === 0 after that, drain the garbage here as no one else will

lib/receiver-queue.js
```diff
@@ -0,0 +1,50 @@
...
+  push (req) {
+    this.queue.push(req)
+    this.requests.set(req.id, req)
+    this.length++
+  }
+
+  shift () {
+    while (this.length > 0) {
+      const req = this.queue.shift()
+      this.requests.delete(req.id)
```
mafintosh, line 19: this should be inside the if also

lib/receiver-queue.js
```diff
@@ -0,0 +1,50 @@
...
+
+    req.block = null
+    req.hash = null
+    req.seek = null
+    req.upgrade = null
+
+    this.requests.delete(id)
+    this.length--
+
+    if (this.length === 0) this._drain()
```
mafintosh, line 42: this.queue.clear()

lib/replicator.js
```diff
@@ -510,6 +511,37 @@ class Peer {
...
+
+  oncancel (msg) {
+    this.receiverQueue.delete(msg.request)
+  }
+
+  async ondrain () {
+    if (this.receiverQueue.length > 0) {
+      await this._handleRequests()
+    }
+  }
```
mafintosh, line 530: could just be
```
ondrain () {
  return this._handleRequests()
}
```

PR #401, opened 2023-06-27
Block requests have higher priority

LuKks:


lib/receiver-queue.js
```diff
@@ -3,19 +3,26 @@ const FIFO = require('fast-fifo')
 module.exports = class ReceiverQueue {
   constructor () {
     this.queue = new FIFO()
+    this.priority = new FIFO()
     this.requests = new Map()
     this.length = 0
```
LuKks, line 8: Looks equal to `requests.size` based on current usage

LuKks: Ah it's not the same, if two requests with same id are pushed

lib/receiver-queue.js
```diff
@@ -3,25 +3,40 @@ const FIFO = require('fast-fifo')
...
   constructor () {
     this.queue = new FIFO()
+    this.priority = []
     this.requests = new Map()
     this.length = 0
   }
 
   push (req) {
-    this.queue.push(req)
+    if (req.priority > 0) this.priority.push(req)
```
mafintosh, line 12: can you add a TODO above this line to use a heap at some point if we wanna support multiple prios

mafintosh:
Looks good!

PR #403, opened 2023-07-03
Add peer.remoteContiguousLength

LuKks:


test/remote-length.js
```diff
@@ -0,0 +1,188 @@
...
+  // await b.get(4)
+  // await b.get(5)
+
+  await new Promise(resolve => setTimeout(resolve, 100))
+
+  // For example, "A" wants to know if "B" is finished syncing
+  const peer = a.replicator.peers.find(peer => peer.remotePublicKey.equals(b.replicator.peers[0].stream.publicKey)) // (unsure if this peers[0] is precise)
+  console.log('B remote contig length', peer.remoteContiguousLength) // => 1
+  console.log('B local contig length ', b.contiguousLength) // => 3
+  t.is(b.contiguousLength, peer.remoteContiguousLength)
```
LuKks, line 32: So `b` got 3 blocks (0, 1, and 2) so `b.contiguousLength` (local contig length) is `3`, that's ok

So now from `a` perspective, it checks the peer `b` but the `peer.remoteContiguousLength` results in `1`

I think there is a bug with the `onrange` callback because it will only set the correct start/length when it's fully downloaded (like shown in the second test)

LuKks: For example, `b` does this:
```js
await b.get(1)
await b.get(2)
await b.get(0)
```

So from `a` perspective, the peer `b` reports this in `onrange`:

```
b2aedb 25ab05 b onrange { drop: false, start: 1, length: 1 }
b2aedb 25ab05 b onrange { drop: false, start: 2, length: 1 }
b2aedb 25ab05 b onrange { drop: false, start: 0, length: 1 }
```

I think I could do some magic with a few variables to try calculate it correctly based on the previous start but I think length not reporting it correctly is probably a bug?

lib/replicator.js
```diff
@@ -920,6 +920,8 @@ class Peer {
   }
 
   _maybeWant (start, length = 1) {
+    if (start + length < this.remoteContiguousLength) return
```
mafintosh, line 923: <=

PR #407, opened 2023-07-26
fix pair bug when core closes

mafintosh:


mafintosh:
should fix https://github.com/holepunchto/corestore/pull/63

PR #413, opened 2023-08-09
pass batch down to sign and verify function

chm-diederichs:


LuKks:
Corestore will stop working due the breaking changes, should we make it backward compatible or update Corestore accordingly?

test/auth.js
```diff
@@ -246,7 +253,7 @@ test('core using custom sign fn', async function (t) {
   t.is(b.length, 5)
 })
 
-function hash (...data) {
+function cryptoHash (...data) {
   const out = b4a.alloc(32)
   sodium.crypto_generichash(out, b4a.concat(data))
```
LuKks, line 258: Sounds like we can remove this helper in favor of `const out = crypto.hash(data)` from hypercore-crypto that we already have installed

LuKks: Unless we need that `alloc` instead of `allocUnsafe` but it should be ok

chm-diederichs:
> Corestore will stop working due the breaking changes, should we make it backward compatible or update Corestore accordingly?

Updated the PR to be backwards compat, now `batch / hash` are passed as the last args

LuKks:
Cool, last thing I think we can safely remove this unused helper now
https://github.com/holepunchto/hypercore/blob/3589b2d18458d388489b12fb9f72ab4222f10fcb/lib/caps.js#L26

chm-diederichs:
These changes we need were merged in #418.

Now this PR would just be removing the `legacy` option, which wasn't the original intention, so I'll close this.

PR #422, opened 2023-08-23
Throw on oplog header overflow

kasperisager:
- [x] https://github.com/holepunchto/hypercore-errors/pull/1

### DIFF
--- a/test/oplog.js
+++ b/test/oplog.js
@@ -358,6 +358,20 @@ test('oplog - multi append is atomic', async function (t) {
   await cleanup(storage)
 })
 
+test('oplog - header overflow', async function (t) {
+  const storage = testStorage()
+
+  const log = new Oplog(storage)
+  await log.open()
+
+  try {
+    await log.flush(Buffer.alloc(8192))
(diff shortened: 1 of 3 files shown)


test/oplog.js
```diff
@@ -358,6 +358,20 @@ test('oplog - multi append is atomic', async function (t) {
...
 })
 
+test('oplog - header overflow', async function (t) {
+  const storage = testStorage()
+
+  const log = new Oplog(storage)
+  await log.open()
+
+  try {
+    await log.flush(Buffer.alloc(8192))
```
mafintosh, line 368: i think 4kb is enough here

mafintosh: (4kb + 1 to make it throw)

kasperisager: The additional 8 bytes used by the oplog currently counts towards the overflow, so it'll throw at 4 KiB - 8 bytes. Is that correct?

mafintosh: ya something like that. might be nice to test the actual limit just to see.

kasperisager:

PR #425, opened 2023-08-28
Add manifest

mafintosh:


lib/core.js
```diff
@@ -725,4 +759,18 @@ function closeAll (...storages) {
   })
 }
 
+function createCompatManifest (publicKey) {
+  return {
+    hash: 'blake2b',
+    namespace: b4a.alloc(32),
```
chm-diederichs, line 765: if you leave `namespace` and `entropy` undefined here, we can handle them when creating the auth

index.js
```diff
@@ -282,26 +282,16 @@ module.exports = class Hypercore extends EventEmitter {
...
-
-    // This only works if the hypercore was fully loaded,
-    // but we only do this to validate the keypair to help catch bugs so yolo
-    if (this.key && keyPair) keyPair.publicKey = this.key
+    const keyPair = opts.keyPair
 
     if (opts.auth) {
       this.auth = opts.auth
-    } else if (opts.sign) {
-      this.auth = Core.createAuth(this.crypto, keyPair, opts)
```
chm-diederichs, line 298: will anything break if we remove this branch?

lib/core.js
```diff
@@ -258,6 +271,10 @@ module.exports = class Core {
     await this.blocks.put(index, value, byteOffset)
   }
 
+  isWritable () {
```
chm-diederichs, line 274: do we use this anymore?

index.js
mafintosh: we tested and corestore doesn't use it so doubt it

lib/core.js
mafintosh: good point

index.js
chm-diederichs: yeh think corestore just passes `auth` so no problem there

mafintosh: 👍

lib/core.js
mafintosh: gone

PR #426, opened 2023-08-29
Parse manifest

chm-diederichs:


lib/manifest.js
```diff
@@ -0,0 +1,200 @@
...
+  }
+}
+
+function signed (signable, signature, idx) {
+  for (let i = 0; i < idx.length; i++) {
+    const indexer = idx[i]
+    if (!indexer.verify(signable, signature)) continue
+
+    const swap = idx.pop()
+    if (indexer !== swap) idx[i] = swap
```
mafintosh, line 194: idx[i--] = swap

index.js
```diff
@@ -74,10 +75,11 @@ module.exports = class Hypercore extends EventEmitter {
     this.writable = false
     this.opened = false
     this.closed = false
+    this.sign = null
```
mafintosh, line 78: dont think this is used

index.js
```diff
@@ -250,15 +261,14 @@ module.exports = class Hypercore extends EventEmitter {
...
-
+    if (!this.keyPair) this.keyPair = o.keyPair
     this.crypto = o.crypto
     this.id = o.id
     this.key = o.key
     this.core = o.core
     this.replicator = o.replicator
     this.encryption = o.encryption
-    this.writable = !this._readonly && !!this.auth && !!this.auth.sign
+    this.writable = !this._readonly || !!(this.keyPair && this.keyPair.secretKey)
```
mafintosh, line 271: should be &&

lib/manifest.js
chm-diederichs: we return immediately after, still good to have it in?

mafintosh: ah ok! yes, nvm then

PR #433, opened 2023-09-12
allow batches to be overtaken in the bg

mafintosh:


chm-diederichs (changes requested):
Perfect, makes sense to just keep in memory if we can't flush.

Only thing is can we return the result from _flush to flush caller

PR #437, opened 2023-09-18
support partial clones

chm-diederichs:


lib/core.js
```diff
@@ -227,7 +227,7 @@ module.exports = class Core {
         const segmentStart = pos
         const segmentEnd = src.bitfield.firstUnset(pos)
 
-        if (segmentStart < 0) break
+        if (segmentStart > length || segmentStart < 0) break
```
mafintosh, line 230: >= ?

lib/core.js
```diff
@@ -217,7 +217,7 @@ module.exports = class Core {
     return false
   }
 
-  async copyFrom (src, signature) {
+  async copyFrom (src, signature, length = src.tree.length) {
```
mafintosh, line 220: make the length an option, { length = src.tree.length } = {}

chm-diederichs:
hold off on merging this, this fixes an autobase bug, which should probably be fixed in autobase

lib/core.js
```diff
@@ -247,7 +247,7 @@ module.exports = class Core {
       // TODO: is it an issue to move the nodes directly?
       // TODO: make flat iterator that computes needed nodes
 
-      for (let i = 0; i <= src.tree.length * 2; i++) {
+      for (let i = 0; i <= length * 2; i++) {
```
mafintosh, line 250: should be `<` not `<=` i think, unrelated

PR #438, opened 2023-09-20
Use b4a in tests

LuKks:


test/encodings.js
```diff
@@ -20,7 +21,7 @@ test('encodings - supports custom encoding', async function (t) {
 test('encodings - supports custom batch encoding', async function (t) {
   const a = await create(null, {
     encodeBatch: batch => {
-      return [Buffer.from(batch.map(b => b.toString()).join('-'))]
```
LuKks, line 23: Removed the `toString` because values are already strings

test/compat.js
```diff
@@ -23,7 +24,7 @@ for (const abi of abis) {
     for (let i = 0; i < 1000; i++) {
       const block = await core.get(i)
 
-      if (!block.equals(Buffer.of(i))) {
+      if (!block.equals(b4a.from([i]))) {
```
mafintosh, line 27: b4a.equals

PR #464, opened 2023-12-13
Support inheriting treeHash from another core

chm-diederichs:


lib/core.js
```diff
@@ -553,6 +561,11 @@ module.exports = class Core {
       const batch = this.tree.batch()
       for (const val of values) batch.append(val)
 
+      // only multisig can have prologue so signature is always present
+      if (this.tree.prologue && !this.verifier.verify(batch, signature)) {
```
mafintosh, line 565: dont think we need this

lib/merkle-tree.js
```diff
@@ -718,7 +723,7 @@ module.exports = class MerkleTree {
       roots.push(await getStoredNode(storage, index, null, true))
     }
 
-    return new MerkleTree(storage, roots, opts.fork || 0, opts.signature || null)
+    return new MerkleTree(storage, roots, opts.fork || 0, opts.signature || null, opts.prologue)
```
mafintosh, line 726: would still || null here for consistentcy

PR #477, opened 2024-01-25
Don't start downloading when core is not active

HDegroote:
Current behaviour doesn't interact well with corestore, because whenever a peer connects on a discovery key with you, this line is called:
https://github.com/holepunchto/corestore/blob/3ab2595c8313179bc98567dbe28389245ddca6e8/index.js#L345

Even though `active: false` is passed in when creating the core, when `this.downloading` is true (which it always is initially) you still end up requesting a download from all the other peers you're connected with on that core, and they end up doing the same, so it propagates.

Note: I'm not sure why `this.downloading` had default true. This is a big behaviour change, so might break something. Fwiw, tests also pass when it's defaulted to false.

### DIFF
--- a/index.js
+++ b/index.js
@@ -396,6 +396,7 @@ module.exports = class Hypercore extends EventEmitter {
       eagerUpgrade: true,
       notDownloadingLinger: opts.notDownloadingLinger,
       allowFork: opts.allowFork !== false,
+      initDownloading: this._active,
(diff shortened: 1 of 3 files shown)


HDegroote:
See https://github.com/holepunchto/corestore/pull/86 for a corestore test illustrating the current behaviour. That test passes when using this PR

HDegroote (commented):
looks good to me

PR #482, opened 2024-03-01
Get undefined block is not allowed

LuKks:
Tests produces different errors but they look related thus same PR

The second test is skipped due high CPU usage, by unskipping it locally you can reproduce

Later, to reproduce the final "Error: Could not load node: 4" the test can be adjusted after the CPU fix or it can be reproduced at the moment by running this script twice:

```js
const Hypercore = require('hypercore')

main()

async function main () {
  const core = new Hypercore('./data-w')
  await core.append(['Hello', 'World']) // Two values is required for this

  const clone = new Hypercore('./data-r', core.key)

  const s1 = core.replicate(true)
  const s2 = clone.replicate(false)

  s1.pipe(s2).pipe(s1)

  await clone.update({ wait: true })

  console.log('0', await clone.get(0))
  console.log('undef', await clone.get(undefined))

  await clone.close()
}
```

When running it for the first time it will use 100% CPU, do CTRL-C, and re-run it again then error should happen

It seems that the writer is only left sensible after the first run because if you remove the .append before the second run then everything appears to be fine but it is not lol, just corruption or whatever waiting to happen

### DIFF
--- a/test/basic.js
+++ b/test/basic.js
@@ -428,3 +428,16 @@ test('append above the max suggested block size', async function (t) {
     t.pass('should throw')
   }
 })
+
+test('get undefined block is not allowed', async function (t) {
+  t.plan(1)
+
+  const core = new Hypercore(RAM)
+
+  try {
+    await core.get(undefined)
+    t.fail()
+  } catch (err) {
+    t.pass(err.code, 'ERR_ASSERTION')
+  }
+})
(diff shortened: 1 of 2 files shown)


mafintosh:
Fixed in 10.33.2 that asserts the uints that are passed to c. Note that passing anything other than uints is undefined behaviour but shouldnt crash your process ever (hence the current asserts)

Mind making this into a proper test? :)

mafintosh:
For people listening, its also not corrupting, its simply throwing on get with an invalid index

mafintosh:
Thanks!

PR #488, opened 2024-03-12
Add getRaw method to Batch API

chm-diederichs:


index.js
```diff
@@ -875,6 +875,8 @@ module.exports = class Hypercore extends EventEmitter {
     let block = await req
     if (!block) return null
 
+    if (opts && opts.raw) return block
```
mafintosh, line 878: so simple!

lib/batch.js
```diff
@@ -170,12 +170,17 @@ module.exports = class HypercoreBatch extends EventEmitter {
...
+  async get (index, opts = {}) {
     if (this.opened === false) await this.opening
     if (this.closing) throw SESSION_CLOSED()
 
     const length = this._sessionLength
-    if (index < length) return this.session.get(index, { ...opts, tree: this._sessionBatch })
+
+    if (index < length) {
+      return this.session.get(index, { ...opts, tree: this._sessionBatch })
+    } else if (opts && opts.raw) {
```
mafintosh, line 181: make two ifs, the else is redundant as you return above

PR #490, opened 2024-03-19
use skip bitfield to optimise range requests further

mafintosh:


lib/replicator.js
```diff
@@ -2218,3 +2240,7 @@ function onwireextension (m, c) {
 function setDownloadingLater (repl, downloading, session) {
   repl.setDownloadingNow(downloading, session)
 }
+
+function isBlockRequest (req) {
+  return req !== null && req.block !== null
```
HDegroote, line 2245: Won't `req.block` be undefined if it's not a block request, since it's either set or not set? (in which case `req.block != null` is better)

mafintosh: no its always explicit (i hate undefined haha)

HDegroote:
Actually, I'm not sure if the current approach skips all inflight blocks. Won't `_findNext`'s early return flow still trigger for inflight blocks? That flow never checks `missingBlocks`, just the bitfields.
https://github.com/holepunchto/hypercore/blob/64116ee4aa66379168b706acb976dd54e1f93f4f/lib/replicator.js#L1062-L1066

mafintosh:
We already track inflight blocks, ie the code is already skipping over them. The issue is to skip over them efficiently

mafintosh:
You are right its prop NOT skipping efficiently in the remote contig case

HDegroote:
Best to hold off on merging this for now, it's throwing errors when I use this as a dependency in a more complex run. They seem unrelated to this PR at first sight, so it might just have uncovered an existing error--will look into it.

HDegroote (approved):
I verified this branch against a scenario with heave download load, and it behaved fine, so this looks good to me

PR #491, opened 2024-03-20
Trace _requestRangBlock and _send start

HDegroote:
Note: _send already has some tracing at the end: https://github.com/holepunchto/hypercore/blob/64116ee4aa66379168b706acb976dd54e1f93f4f/lib/replicator.js#L1211
But there are some early-return flows, so I think it makes sense to trace the beginning too

### DIFF
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -1034,6 +1034,7 @@ class Peer {
   }
 
   _requestRangeBlock (index, length) {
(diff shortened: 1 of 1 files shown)


HDegroote:
These would serve more for measuring performance rather than tracing, and involve private methods which can be refactored at any time, so I'll close this

PR #493, opened 2024-03-22
Expose inflightRange opt

HDegroote:
Useful for example when downloading many cores from many peers, where specifying a low fixed number of inflight blocks can turn out to be more efficient

### DIFF
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -1242,8 +1242,7 @@ module.exports = class Replicator {
     this.downloading = false
     this.activeSessions = 0
 
-    this.peerOpts = {}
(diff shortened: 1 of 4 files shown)


lib/replicator.js
```diff
@@ -2090,7 +2093,8 @@ module.exports = class Replicator {
 
     if (channel === null) return onnochannel()
 
-    const peer = new Peer(replicator, protomux, channel, session)
+    const opts = { inflightRange: this.inflightRange }
+    const peer = new Peer(replicator, protomux, channel, session, opts)
```
mafintosh, line 2097: this is an internal object so just pass the inflightRange as an argument. its not optional here in practice

PR #496, opened 2024-03-26
set lower bound of inflights to 16 and factor in total peers in math …

mafintosh:
…also

### DIFF
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -33,7 +33,7 @@ const m = require('./messages')
 const caps = require('./caps')
 const { createTracer } = require('hypertrace')
 
-const DEFAULT_MAX_INFLIGHT = [32, 512]
+const DEFAULT_MAX_INFLIGHT = [16, 512]
 const SCALE_LATENCY = 50
 const DEFAULT_SEGMENT_SIZE = 256 * 1024 * 8 // 256 KiB in bits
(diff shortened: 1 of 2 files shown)


lib/replicator.js
```diff
@@ -369,7 +369,7 @@ class Peer {
     const stream = this.stream.rawStream
     if (!stream.udx) return Math.min(this.inflightRange[1], this.inflightRange[0] * 3)
 
-    const scale = stream.rtt <= SCALE_LATENCY ? 1 : stream.rtt / SCALE_LATENCY
+    const scale = stream.rtt <= SCALE_LATENCY ? 1 : stream.rtt / SCALE_LATENCY * Math.min(1, 2 / this.replicator.peers.length)
```
HDegroote, line 372: The result can now be less than  the minimum of `DEFAULT_MAX_INFLIGHT` (16 by default), which I don't think is desired. I think the contract of this function is that it always returns a value in `[16, 512]` (or whatever the `DEFAULT_MAX_INFLIGHT` values are).

Consider for example the case of 100 peers and `stream.rtt` 100.
Scale: `100 / 50 * 2  / 100 = 4/100 = 0.04`
Result: `Min(512, 16*0.04) = 0.64`

=> If negative scaling is desired, then I think it should be applied to the max inflight range, with a sanity check of returning at least the min inflight range.

PR #497, opened 2024-03-26
Pass inflightRange on to new session

HDegroote:
Note: I could do the same for the 'clone' method, which currently doesn't pass on `inflightRange`. Unsure if that makes sense though (I've never used `clone`)


mafintosh:
tweaked in main, so closing

PR #502, opened 2024-03-30
chore: Remove unused `length` arg in `bitfield.getBitfield()`

lejeunerenard:
Looks like it was never used. Likely accidentally included.


mafintosh:
Thanks!

PR #504, opened 2024-04-08
chore: add missing docs for upload and download events

gmaclennan:
These events are not currently documented, but I assume they are officially supported? I think these docs are correct, but please review!


mafintosh:
Yes and thanks!

PR #507, opened 2024-04-19
do not allocate a full session for replication sessions

mafintosh:


index.js
```diff
@@ -250,12 +250,16 @@ module.exports = class Hypercore extends EventEmitter {
...
-
-    this.sessions.push(s)
+    this._addSession(s)
 
     return s
   }
 
+  _addSession (s) {
+    this.sessions.push(s)
+    if (this.core) this.core.active++
```
HDegroote, line 260: I'm not seeing how `this.core` can be unset. The book-keeping code becomes much easier to reason about without that `if`. So if it's indeed unneeded, better to remove it

 `_addSession ` is called either from `_openFromExisting(...)`, or in `session(...)`

In `session(...)` it runs after `_passCapabilities`, which sets `this.core`

In `_openFromExisting` it also runs after `_passCapabilities

index.js
```diff
@@ -479,14 +487,14 @@ module.exports = class Hypercore extends EventEmitter {
...
       this.replicator.findingPeers -= this._findingPeers
       this.replicator.clearRequests(this.activeRequests, err)
-      this.replicator.updateActivity(this._active ? -1 : 0, this)
+      this.replicator.updateActivity(this._active ? -1 : 0)
     }
 
     this._findingPeers = 0
 
-    if (this.sessions.length) {
+    if (this.sessions.length || this.core.active > 0) {
```
HDegroote, line 495: Sanity check on the intent of the flow:
It's possible for `this.core.active`to be > 0, while `this.sessions.length` is 0, right? And in that case we can just close without worrying about closing any sessions (since there are none).

index.js
```diff
@@ -479,14 +487,14 @@ module.exports = class Hypercore extends EventEmitter {
...
       // if this is the last session and we are auto closing, trigger that first to enforce error handling
-      if (this.sessions.length === 1 && this.autoClose) await this.sessions[0].close(err)
+      if (this.sessions.length === 1 && this.core.active === 1 && this.autoClose) await this.sessions[0].close(err)
```
HDegroote, line 497: It's not clear to me how the case where `this.core.active > 1` and `this.sessions.length === 1` is handled (or if it can occur). In that case, is that session guaranteed to be closed later?

test/replicate.js
```diff
@@ -1325,51 +1333,6 @@ test('idle replication sessions auto gc', async function (t) {
   t.ok(closed, 'replication session gced')
 })
 
-test('idle replication sessions auto gc with timing', async function (t) {
```
HDegroote, line 1328: Sanity check: has this test become irrelevant now? Seems like it could be relevant for detecting issues with this PR if it's possible to adapt it

test/sessions.js
```diff
@@ -51,7 +51,6 @@ test('sessions - auto close', async function (t) {
 
   await a.close()
   t.absent(closed, 'not closed yet')
-
```
HDegroote, line 54: Detail: unneeded change

index.js
mafintosh: correct, active is the "correct" number always

lib/replicator.js
```diff
@@ -2086,16 +2086,27 @@ module.exports = class Replicator {
     this._maybeResolveIfAvailableRanges()
   }
 
-  _closeSession (session) {
-    session.close().catch(safetyCatch)
+  _closeSession () {
+    this.core.active--
+    if (this.core.sessions.length === 1 && this.core.active === 1 && this.core.sessions[0].autoClose) {
```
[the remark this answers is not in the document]
mafintosh: yes it does! its some gnarly logic due to corestore (needs a refactor). this just checks if there is only one session left and thats the only active holder to the core, and if so if that session wants to be auto closed

test/replicate.js
mafintosh: ya just too many private things being inspected here so killed it, you'll need to hook into peer._closeSession for this now

mafintosh: or mb just on('peer-remove')

index.js
HDegroote: Okay, makes sense (assuming you meant `replicator.js` instead of core.js, but 99% sure you did)

lib/replicator.js
```diff
@@ -2086,16 +2086,27 @@ module.exports = class Replicator {
...
+      this.core.sessions[0].close().catch(safetyCatch)
+    }
+    if (this.core.active === 0 && this.core.sessions.length === 0) {
```
HDegroote, line 2094: Could benefit from a comment; because I'm 90% sure it will take me a while to understand this code next time I read it. Something like

`// the underlying session already closed before, and this is the last replication session. So the core should be closed now.

index.js
```diff
@@ -465,6 +472,7 @@ module.exports = class Hypercore extends EventEmitter {
     if (i === -1) return
 
     this.sessions.splice(i, 1)
+    this.core.active--
```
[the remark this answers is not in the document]
mafintosh: ya that should never happen

lib/replicator.js
```diff
@@ -2086,16 +2086,32 @@ module.exports = class Replicator {
...
+
+    // we were the last active ref, so lets shut things down
+    if (this.core.active === 0 && this.core.sessions.length === 0) {
+      this.destroy()
+      this.core.close().catch(safetyCatch)
+      return
+    }
+
+    // in case one session is still alive but its been marked for auto close also kill it
+    if (this.core.sessions.length === 1 && this.core.active === 1 && this.core.sessions[0].autoClose) {
```
HDegroote, line 2100: 2 final sanity checks:
- Can the case `(this.core.active === 1 && this.core.sessions.length === 0)` or `this.core.active === 0 && this.core.sessions.length === 1` ever happen? (we only consider 0-0 and 1-1)
- is it relevant to consider the case of multiple remaining sessions, all marked for auto close?

mafintosh: only one autoclose one can exist (its a corestore option)

mafintosh: Ya active can easily be 1 or more here with no explicit sessions. That just means more replication sessions exist which will then go through this code later

HDegroote (approved):

PR #508, opened 2024-04-21
add tmp quickAudit api

mafintosh:


lib/core.js
```diff
@@ -194,6 +194,88 @@ module.exports = class Core {
...
+          }
+        }
+
+        i++
+      }
+    } finally {
+      await this._mutex.unlock()
+    }
+
+    if (drops) await this.flush()
```
HDegroote, line 275: Can anything bad happen if an operation runs between the `await this._mutex.unlock()` just above and the `this.flush()`? There's no guarantee such an operation wouldn't access the mutex.

HDegroote: Couldn't the same result as calling `this.flush()` be achieved within the mutex lock, just before closing the `try`? `flush` just does

```
      this._manifestFlushed = true
      this._autoFlush = 4
      await this._flushOplog()
```
I'm not sure if the first 2 statements are relevant, so perhaps just calling `await this._flushOplog()` suffices

mafintosh: it should be fine, its just not in try since it also requires the mutex.

lib/core.js
```diff
@@ -283,6 +297,11 @@ module.exports = class Core {
           start: segmentStart,
           length: segmentEnd - segmentStart
         })
+        console.log({
```
HDegroote, line 300: Left over debug statement?

lib/core.js
```diff
@@ -194,6 +195,19 @@ module.exports = class Core {
     return new this(header, compat, crypto, oplog, bigHeader, tree, blocks, bitfield, verifier, legacy, opts.onupdate || noop, opts.onconflict || noop)
   }
 
+  async audit () {
+    await this._mutex.lock()
+
+    try {
+      await this._flushOplog()
```
HDegroote, line 202: Not a comment, but a question: in which case does the oplog need flushing here, or is it more of a just-in-case sanity check?

lib/audit.js
```diff
@@ -0,0 +1,120 @@
...
+      corrections.tree++
+    }
+  }
+}
+
+function readFullStorage (storage) {
+  return new Promise((resolve) => {
+    storage.stat((_, st) => {
+      if (!st) return resolve(b4a.alloc(0))
+      storage.read(0, st.size, (_, data) => resolve(data || null))
```
HDegroote, line 117: The code in `auditCore` assumes a buffer is returned (it would crash if `null` is returned).

If the read from the storage fails (unlikely, but possible I think), crashing as loud as possible seems like a saner default, just in case it's a temporary failure, because then we break the invariant that the hypercore is fixed after `auditCore`finishes

lib/audit.js
```diff
@@ -0,0 +1,120 @@
...
+      if (i < 0) break
+      nextOffset = -1
+      continue
+    }
+
+    if (nextOffset === -1) {
+      try {
+        nextOffset = await core.tree.byteOffset(i * 2)
+      } catch {
+        core._setBitfield(i, false)
```
HDegroote, line 72: If a particular error is expected, it's more conservative to check for it explicitly, to avoid any risk of removing a bitfield that was fine.

lib/core.js
mafintosh: cause we load it all from the files in audit

lib/audit.js
mafintosh: it should never throw basically (outside storage failure)

HDegroote: Yeah, that's the only edge case I could immediately think of besides a syntax error sneaking into `byteOffset` undetected by the tests somehow.

If it only throws on edge cases, any sense in removing the try-catch and just making it crash on error? But fine with as-is solution too, it does seem very unlikely for the storage to fail just here

lib/core.js
mafintosh: uf, yes

lib/audit.js
mafintosh: throw now

lib/audit.js
mafintosh: (it will also throw on corruption so thats why its there)

PR #510, opened 2024-04-25
copyFrom should not be destructive

chm-diederichs:


lib/merkle-tree.js
```diff
@@ -824,7 +824,7 @@ function verifyUpgrade ({ fork, upgrade }, blockRoot, batch) {
 
   if (prologue) {
     const { start, length } = upgrade
-    if (start < prologue.length && (start !== 0 || length !== prologue.length)) {
+    if (start < prologue.length && (start !== 0 || length < prologue.length)) {
```
chm-diederichs, line 827: note this change here, allows any upgrade >= prologue length

chm-diederichs: before was just === prolouge.length

chm-diederichs: i think it was a typo, but just double checking there isn't a reason for that

PR #512, opened 2024-04-30
add failing batch flush test

mafintosh:


mafintosh:
deprecating the pending one so this can close

PR #513, opened 2024-04-30
Don't send unneeded 'wireRange' messages

HDegroote:
Currently `broadcastRange` sends a `wireRange` message to every peer on every update. This PR avoids sending this message to peers when we already know they have those blocks available.

<strike>Failing tests (skipped for now) indicate that we rely on the `broadcastRange` spam to update the `contiguousLength`, so I marked this as draft until that's clarified.</strike>


HDegroote:
A potential extension is to also avoid sending `wireRange` messages if we know the other peer is not downloading (`this.remoteDownloading === false`), but that's relying on different logic so I haven't included it in this PR

lib/replicator.js
```diff
@@ -386,6 +386,21 @@ class Peer {
     if (drop) this._unclearLocalRange(start, length)
     else this._clearLocalRange(start, length)
 
+    // TODO: drop===true case
```
HDegroote, line 389: I think it's fine to ignore the `drop===true` case for now, and have it behave as before. We can then iterate on that later if we notice it becomes relevant

HDegroote:
The tests were no longer relevant: they were checking the writer peer's perception of the downloading peers' contiguous length. I wrote some new tests for the current behaviour on broadcasting the contiguouslength, from the point of view of reader peers.

lib/replicator.js
```diff
@@ -386,6 +386,20 @@ class Peer {
...
+    // TODO: consider also adding early-returns on the drop===true case
+    if (!drop) {
+      if (this._remoteContiguousLength >= start + length) return
+
+      // TODO: consider doing more efficiently. Maybe a function directly in the bitfield logic?
+      let nrFound = 0
+      for (let i = start; i < start + length; i++) {
+        if (!this.remoteBitfield.get(i)) break
+        nrFound++
+      }
```
mafintosh, line 398: just search for the next false (there is an fast api for that). if the next false is < start+length, send, otherwise no

HDegroote:
I think the logic might be flawed, marking as draft while I debug some more

HDegroote:
False alarm, what I detected was unrelated to this PR (fixed here https://github.com/holepunchto/hypercore/pull/517)

PR #517, opened 2024-05-06
Fix RemoteBitfield findFirst edge case 

HDegroote:


test/remote-bitfield.js
```diff
@@ -27,3 +30,41 @@ test('set last bits in segment and findFirst', function (t) {
   t.is(b.findFirst(false, 32766), 32768)
   t.is(b.findFirst(false, 32767), 32768)
 })
+
+test('remote congituous length consistency (remote-bitfield findFirst edge case)', async function (t) {
```
HDegroote, line 34: I added the test to the remote bitfield tests because it detects an issue with the bitfield logic, but since it's using the replicator API it could also live in `test/remote-length.js`

PR #518, opened 2024-05-09
added test for session id reuse

rafapaezbas:


test/replicate.js
```diff
@@ -1470,6 +1470,43 @@ test('can define default max-inflight blocks for replicator peers', async functi
...
+  a.replicate(n1)
+  b.replicate(n2)
+
+  const blocks = Array.from(Array(n).keys())
+
+  while (blocks.length > 0) {
+    const session = await b.session()
+    for (let i = 0; i < 100; i++) {
+      if (blocks[i]) {
+        session.get(i).catch((err) => { return err })
```
mafintosh, line 1492: session.get(i).catch(noop) is fine where noop is function noop () {}

test/replicate.js
```diff
@@ -1470,6 +1470,43 @@ test('can define default max-inflight blocks for replicator peers', async functi
...
 })
 
+test('session id reuse does not stall', async function (t) {
+  t.plan(1)
+  const a = await create()
+  const b = await create(a.key)
+
+  const n = 5000
+
+  for (let i = 0; i < n; i++) await a.append(b4a.from([0]))
```
mafintosh, line 1480: make one batch instead, ie bill up an array with 5k blocks and pass that to array. much faster

PR #521, opened 2024-05-27
Copy node-cache entries to reduce mem usage

HDegroote:
Context: random-access-file uses slab-allocated memory by default when reading from disk. Since we store the nodes in a cache, the entire slab is kept from being garbage-collected. This PR copies the cache entry to a new, smaller slab.

### DIFF
--- a/lib/merkle-tree.js
+++ b/lib/merkle-tree.js
@@ -546,6 +546,7 @@ module.exports = class MerkleTree {
     if (c) return c
 
     let node = this.unflushed.get(index)
+
     if (this.flushing !== null && node === undefined) {
       node = this.flushing.get(index)
     }
--- a/lib/merkle-tree.js
+++ b/lib/merkle-tree.js
@@ -1206,18 +1206,15 @@ function blankNode (index) {
 // Storage methods
 
 function getStoredNode (storage, index, cache, error) {
-  const nodeSize = 40
-
   return new Promise((resolve, reject) => {
-    storage.read(nodeSize * index, nodeSize, (err, data) => {
+    storage.read(40 * index, 40, (err, data) => {
       if (err) {
         if (error) return reject(err)
         else resolve(null)
         return
       }
 
-      const uintNrBytes = 8
-      const hash = data.subarray(uintNrBytes)
+      const hash = data.subarray(8)
       const size = c.decode(c.uint64, data)
 
       if (size === 0 && b4a.compare(hash, BLANK_HASH) === 0) {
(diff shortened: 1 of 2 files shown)


lib/merkle-tree.js
```diff
@@ -1224,7 +1227,25 @@ function getStoredNode (storage, index, cache, error) {
       }
 
       const node = { index, size, hash }
-      if (cache !== null) cache.set(index, node)
+
+      if (cache !== null) {
```
HDegroote, line 1231: Note: the logic could be simplified if we always copy the node's buffers (not only when cache is enabled). I think that would make sense too

lib/merkle-tree.js
```diff
@@ -1224,7 +1227,25 @@ function getStoredNode (storage, index, cache, error) {
...
+        const slab = b4a.alloc(nodeSize)
+        b4a.copy(data, slab)
+
+        const sizeCopy = c.decode(
```
HDegroote, line 1235: I'm not sure if it's needed to copy the size. When decoding a buffer into an integer, is the original buffer still referenced? (if not, only the hash needs to be copied)

HDegroote:
Marking as draft since there is a failing CI on windows I need to look into (seems like an interaction with tmp-test)

lib/merkle-tree.js
mafintosh: we can just do this:

```js
const node = .... // like before

if (cache !== null) {
  node.hash = b4a.from(node.hash) // copy
  cache.set(index, node)
}
```

as the hash is the only thing keeping the slab alive

HDegroote: Nice, not having to copy the int makes it easier.

`b4a.from(...)` does seem to use slab memory though, the test shows its underlying buffer has length ~8kb, so I just explicitly copied the hash (and nothing else)

lib/merkle-tree.js
```diff
@@ -1224,7 +1224,15 @@ function getStoredNode (storage, index, cache, error) {
       }
 
       const node = { index, size, hash }
-      if (cache !== null) cache.set(index, node)
+
+      if (cache !== null) {
+        const hashCopy = b4a.alloc(hash.byteLength)
+        b4a.copy(hash, hashCopy)
```
mafintosh, line 1230: you can just use `hashCopy.set(hash)`

lib/merkle-tree.js
```diff
@@ -1224,7 +1224,15 @@ function getStoredNode (storage, index, cache, error) {
       }
 
       const node = { index, size, hash }
-      if (cache !== null) cache.set(index, node)
+
+      if (cache !== null) {
+        const hashCopy = b4a.alloc(hash.byteLength)
```
mafintosh, line 1229: add a one line comment about why we do the copy also for future hans and mathias

PR #524, opened 2024-05-30
Ensure roundtrip request ids are requeued correctly 

chm-diederichs:


lib/replicator.js
```diff
@@ -300,9 +300,15 @@ class RoundtripQueue {
   }
 
   clear () {
-    const q = this.queue
+    const ids = new Array(this.queue.length)
+    for (let i = 0; i < ids.length) {
+      const [, id] = this.queue[i]
+      ids[i] = id
```
mafintosh, line 306: `ids[i] = this.queue[i][1]`

PR #526, opened 2024-06-17
`rocksdb` : add byte helpers and use atomic batches

chm-diederichs:


lib/merkle-tree.js
```diff
@@ -221,24 +222,28 @@ class MerkleTreeBatch {
...
+    const batch = this.tree.storage.createWriteBatch()
     if (this.ancestors < this.treeLength) {
       if (this.ancestors > 0) {
         const head = 2 * this.ancestors
         const ite = flat.iterator(head - 2)
 
         while (true) {
           if (ite.contains(head) && ite.index < head) {
-            this.tree.unflushed.set(ite.index, blankNode(ite.index))
+            batch.addTreeNode(blankNode(ite.index))
```
mafintosh, line 246: now that we have rocks we can just delete nodes! Can you add that api for tree nodes to storage

lib/merkle-tree.js
```diff
@@ -219,40 +213,30 @@ class MerkleTreeBatch {
...
-
-        while (true) {
-          if (ite.contains(head) && ite.index < head) {
-            this.tree.unflushed.set(ite.index, blankNode(ite.index))
-          }
-          if (ite.offset === 0) break
-          ite.parent()
-        }
-      }
+      dbBatch.deleteTreeNodeRange(this.batchId, this.ancestors, this.treeLengthq)
```
mafintosh, line 233: you also need to queue the nodes that are gaps now. thats what the code on the left here does, but instead of setting blank nodes you can just queue a deletion obvs

ie
   
```
     3
 1      5
0 2  4 6
```

and you truncate 6 out, then you need to remove 3 also

mafintosh (approved):

PR #534, opened 2024-07-02
never send an invalid block request

mafintosh:
Really hard to encapsulate in a test case unfortunately but there is a bug where if we can a range update before a sync update, then we might request an OOB block (which crashes it atm)

This fixes the crash, but also makes sure to not send invalid requests in the first place


HDegroote (approved):
Looks good to me

PR #541, opened 2024-07-24
Unslab downloaded blocks

HDegroote:
Currently, downloaded blocks retain their slab (8kb, or 65kb if udx is used).

<strike>If the hypercore cache is activated, that takes up a lot of memory. But I think it makes sense to always unslab the blocks, regardless of whether they are cached (this makes the fix much easier). </strike>
We ended up unslabbing only in cacheOnResolve. Note that this slightly changes the caching behaviour in case of a parallel first get (see the modified test)

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ const z32 = require('z32')
 const id = require('hypercore-id-encoding')
 const safetyCatch = require('safety-catch')
 const { createTracer } = require('hypertrace')
+const unslab = require('unslab')
 
 const Replicator = require('./lib/replicator')
 const Core = require('./lib/core')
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -911,7 +912,12 @@ module.exports = class Hypercore extends EventEmitter {
   }
 
   async _cacheOnResolve (index, req, fork) {
-    const block = await req
+    const resolved = await req
+
+    // Unslab only when it takes up less then half the slab
+    const block = resolved !== null && 2 * resolved.length < resolved.buffer.byteLength
```
mafintosh, line 918: resolve.byteLength

PR #551, opened 2024-08-13
Unslab userdata values and all header buffers

HDegroote:
Note: the userdata value is set through (at least) 2 code paths:
- through `updateUserData()`
- through the `header` value of `oplog.open()` (in `resume()`)
So I unslabbed it in both

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -100,6 +100,7 @@ module.exports = class Core {
       }
 
       const keyPair = opts.keyPair || (opts.key ? null : crypto.keyPair())
+
       const defaultManifest = !opts.manifest && (!!opts.compat || !opts.key || !!(keyPair && b4a.equals(opts.key, keyPair.publicKey)))
       const manifest = defaultManifest ? Verifier.defaultSignerManifest(opts.key || keyPair.publicKey) : Verifier.createManifest(opts.manifest)
 
@@ -121,6 +122,10 @@ module.exports = class Core {
         }
       }
 
+      if (header.keyPair && !header.keyPair.secretKey) {
+        header.keyPair.secretKey = null
+      }
+
       await flushHeader(oplog, bigHeader, header)
     } else if (header.external) {
       header = await bigHeader.load(header.external)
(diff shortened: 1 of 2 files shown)


lib/core.js
```diff
@@ -1022,6 +1029,8 @@ function addReorgHint (list, tree, batch) {
 }
 
 function updateUserData (list, key, value) {
+  value = unslab(value)
```
mafintosh, line 1032: defer this to after the validation

lib/core.js
```diff
@@ -131,6 +131,9 @@ module.exports = class Core {
     header.tree.rootHash = unslab(header.tree.rootHash)
     header.tree.signature = unslab(header.tree.signature)
 
+    if (header.keyPair?.publicKey) header.keyPair.publicKey = unslab(header.keyPair.publicKey)
+    if (header.keyPair?.secretKey) header.keyPair.secretKey = unslab(header.keyPair.secretKey)
```
mafintosh, line 135: if (header.keyPair) {
  header.keyPair.publicKey = unslab(...)
  header.keyPair.secretKey = unslab(...)
}

HDegroote: That was my initial approach, but that doesn't work for the case where keyPair only contains a publicKey, and no secretKey

mafintosh: publicKey is not null?

mafintosh: if thats undefined and on the header, thats a bug

HDegroote: Ok, there's a test case which hits that, so that's a bug, looking into it

lib/core.js
HDegroote: You mean deferring to when `value` is actually assigned to something (lines 1036 and 1041)? 
(not sure where validation happens)

mafintosh: ah i misremembered, this is ok

lib/core.js
HDegroote: Solved by setting secretKey to null here is it's not set: https://github.com/holepunchto/hypercore/pull/551/files#diff-826e5ed9069c92baa50ec46dde8074a6789b4c90837c5ede0e6b0438e3c34309R124-R126

lib/core.js
```diff
@@ -121,6 +121,10 @@ module.exports = class Core {
         }
       }
 
+      if (header.keyPair && !header.keyPair.secretKey) {
+        header.keyPair.secretKey = null
+      }
```
mafintosh, line 126: just do it up above when the header is constructed, so we dont mutate a users config object,

```
keyPair: keyPair ? { publicKey: keyPair.publicKey, secretKey: keyPair.secretKey || null} : null
```

HDegroote: Good point, fixed

PR #553, opened 2024-08-21
Add upload, download and append stats

HDegroote:
<strike>The way of passing the stats object to all sessions is pretty ad-hoc--if there's a cleaner, more explicit way, that would be preferable,</strike>

stats now live in Core instead

### DIFF
--- a/index.js
+++ b/index.js
@@ -606,7 +606,16 @@ module.exports = class Hypercore extends EventEmitter {
   }
 
   get stats () {
-    return this.core.stats
+    return this.core !== null
+      ? this.core.stats
+      : {
+          blocksUploaded: 0,
+          blocksDownloaded: 0,
+          bytesUploaded: 0,
+          bytesDownloaded: 0,
+          blocksAppended: 0,
+          bytesAppended: 0
+        }
   }
 
   ready () {
--- a/test/replicate.js
+++ b/test/replicate.js
@@ -1630,7 +1630,20 @@ test('seek against non sparse peer', async function (t) {
 })
 
 test('hypercore stats', async function (t) {
-  const a = await create()
+  const a = new Hypercore(RAM)
+  t.alike(
+    a.stats,
+    {
+      blocksUploaded: 0,
+      blocksDownloaded: 0,
+      bytesUploaded: 0,
+      bytesDownloaded: 0,
+      blocksAppended: 0,
+      bytesAppended: 0
+    },
+    'all-0 stats before ready'
+  )
+
   await a.append(['block0', 'block1', 'block2'])
 
   const b = await create(a.key)


index.js
```diff
@@ -256,7 +267,12 @@ module.exports = class Hypercore extends EventEmitter {
   }
 
   _addSession (s) {
+    if (this.sessions.lenth >= 1) {
```
HDegroote, line 270: I think it's guaranteed to always contain at least 1 element, but not 100% sure that's true, so I added the if

HDegroote: No longer relevant

HDegroote:
Marking this as draft--it needs an iteration on where these stats should live exactly

index.js
```diff
@@ -605,13 +605,29 @@ module.exports = class Hypercore extends EventEmitter {
...
+
   ready () {
     return this.opening
   }
 
   _onupload (index, value, from) {
     const byteLength = value.byteLength - this.padding
 
+    this.core.stats.blocksUploaded++
+    this.core.stats.bytesUploaded += byteLength
```
mafintosh, line 629: should not live in the sesison, move to replicator or who ever is triggering this

index.js
```diff
@@ -711,6 +727,9 @@ module.exports = class Hypercore extends EventEmitter {
     if (value) {
       const byteLength = value.byteLength - this.padding
 
+      this.core.stats.blocksDownloaded++
+      this.core.stats.bytesDownloaded += byteLength
```
mafintosh, line 731: same as above

index.js
```diff
@@ -1021,12 +1040,18 @@ module.exports = class Hypercore extends EventEmitter {
...
+    let totalByteLength = 0
     for (const b of buffers) {
       if (b.byteLength > MAX_SUGGESTED_BLOCK_SIZE) {
         throw BAD_ARGUMENT('Appended block exceeds the maximum suggested block size')
       }
+      totalByteLength += b.byteLength
     }
 
+    this.core.stats.blocksAppended += blocks.length
+    this.core.stats.bytesAppended += totalByteLength
```
mafintosh, line 1053: move to core.append instead of the session

index.js
HDegroote: This is emitted from the replicator, but the replicator has no direct access to the `byteLength` as calculated here, so I need to look into another way to calculate that (from what I can tell, it has no access to info about the encryption and its padding, but it can probably know the byte length of the block by checking Hypercore.core, which it does have access to).

mafintosh: byteLength is value.byteLength, this is actualy wrong from a stats pov

mafintosh: the padding is due to the session being encrypted, nothing to do with the size of the block it pulled from the network

HDegroote:
Closing, since this is done more cleanly in
https://github.com/holepunchto/hypercore/pull/553
https://github.com/holepunchto/hypercore/pull/557

PR #557, opened 2024-09-03
Add blocks/bytes uploaded/downloaded stats

HDegroote:
If you prefer, we can also use
```
blocks: { uploaded: 0, downloaded: 0 }
```
Instead of
```
blocksUploaded: 0,
blocksDownloaded: 0
```

But I found that looked a bit strange (particularly for `bytes`)

### DIFF
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -367,7 +367,11 @@ class Peer {
       wireWant: { tx: 0, rx: 0 },
       wireBitfield: { tx: 0, rx: 0 },
       wireRange: { tx: 0, rx: 0 },
-      wireExtension: { tx: 0, rx: 0 }
+      wireExtension: { tx: 0, rx: 0 },
+      blocksUploaded: 0,
+      blocksDownloaded: 0,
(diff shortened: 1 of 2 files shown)


lib/replicator.js
```diff
@@ -2004,6 +2017,10 @@ module.exports = class Replicator {
 
   _ondata (peer, req, data) {
     if (data.block !== null) {
+      this.stats.blocksDownloaded++
```
HDegroote, line 2020: Please double check that this code runs if-and-only-if a block was downloaded. I think this is the correct shared path, but not 100% sure

lib/replicator.js
```diff
@@ -2004,6 +2017,10 @@ module.exports = class Replicator {
...
+      this.stats.bytesDownloaded += data.block.value.byteLength
```
HDegroote, line 2021: Note: this will crash if a block ever does not contain a Buffer `block.value` entry. The code assumes this never happens (which I think is true, but good to confirm)

PR #562, opened 2024-09-06
rocksdb: add db snapshots

chm-diederichs:


lib/core.js
```diff
@@ -124,29 +124,57 @@ class SessionState {
...
     this.treeLength = treeLength
+    this.compatLength = treeLength
+
     this.active = 1
 
     this._onflush = null
     this._flushing = null
     this._activeBatch = null
+
+    this._snapshot = snapshot || null
```
mafintosh, line 135: can we name this storageSnapshot and have it be next to the storage? then it can use it in the stream to make the stream mega fast

PR #568, opened 2024-09-16
Fix activity count

mafintosh:


lib/replicator.js
```diff
@@ -537,7 +537,7 @@ class Peer {
     const reopen = isRemote === true && this.remoteOpened === true && this.remoteDownloading === false &&
        this.remoteUploading === true && this.replicator.downloading === true
 
-    if (this.useSession && !reopen) this.replicator._closeSession()
+    if (this.useSession && !reopen) this.replicator._closeSessionMaybe()
```
HDegroote, line 540: Is the intent of this line to ensure the session is closed when `this.remoteOpened === false`? If so, I'd place it inside the if on line 542, to make it obvious. (For the case where the if doesn't trigger, the `this.replicator.removePeer`  already calls `closeSessionMaybe`, although the if there only checks for `this.useSession`, not caring about `reopen`

lib/replicator.js
```diff
@@ -2275,12 +2295,15 @@ module.exports = class Replicator {
     protomux.stream.setMaxListeners(0)
     protomux.stream.on('close', this._onstreamclose)
 
+    if (useSession) this._peerSessions++
```
HDegroote, line 2298: Detail: what do you think of adding this line just before `ensureSession`? I think those 2 statements are related, so that makes it clearer (I think it makes no sense to call `this.ensureSession` without incrementing `this._peerSessions`, so we could also put it before `this.ensureSession()`. That way there's an invariant: 'ensureSession is only ever called if this._peerSessions > 0')

lib/replicator.js
```diff
@@ -2275,12 +2295,15 @@ module.exports = class Replicator {
...
     this._ifAvailable++
+
     protomux.stream.opened.then((opened) => {
+      if (useSession) this._peerSessions--
       this._ifAvailable--
 
       if (opened && !this.destroyed) makePeer()
```
HDegroote, line 2305: Sanity check: no need to check `this.core.closed` here?

lib/replicator.js
mafintosh: no it should run everytime the channel closes, except when we immediately trigger a reopen

lib/replicator.js
mafintosh: no its related to the line below, ie, its to avoid the core closing underneath we dbl increment here.

lib/replicator.js
mafintosh: prop should be unrelated to this i think

PR #578, opened 2024-10-04
rocksdb: only create single read batch in bg call

chm-diederichs:


lib/replicator.js
```diff
@@ -1838,15 +1838,20 @@ module.exports = class Replicator {
...
+    const blocks = new Map()
     for (const b of this._blocks) {
-      if (this.core.bitfield.get(b.index) === false) continue
+      blocks.set(b.index, this.core.blocks.get(reader, b.index))
+    }
 
-      const reader = this.core.storage.createReadBatch()
+    reader.tryFlush()
+
+    for (const b of this._blocks) {
```
HDegroote, line 1850: Sanity check: this shouldn't be `for (const b of blocks) {` ? Otherwise it's iterating over a data structure than can mutate during iteration, so line 1851 can have pretty strange behaviour if the entry doesn't exist.

lib/replicator.js
```diff
@@ -1838,15 +1838,20 @@ module.exports = class Replicator {
 
     let clear = null
 
+    const reader = this.core.storage.createReadBatch()
+
+    const blocks = new Map()
     for (const b of this._blocks) {
-      if (this.core.bitfield.get(b.index) === false) continue
+      blocks.set(b.index, this.core.blocks.get(reader, b.index))
```
HDegroote, line 1845: Sanity check: `this.core.blocks.get(reader, b.index)` can never reject? Otherwise I think it needs a safetyCatch or noop catch-handler, since there are async operations before the promises are awaited

HDegroote (approved):
LGTM (with the caveat that I don't understand the read-batch logic, so I didn't check that)

mafintosh (approved):

PR #580, opened 2024-10-10
rocksdb: move `MemoryOverlay` to storage module

chm-diederichs:


chm-diederichs:
depends on holepunchto/hypercore-on-the-rocks#25

PR #583, opened 2024-10-15
Add failing batch test

mafintosh:


mafintosh:
should be fixed by https://github.com/holepunchto/hypercore-storage/pull/28

mafintosh:
In addition to this, see the TODO in the storage PR. Instead of enforcing the existance of the head record (implies the tree hash is optional which is weird), we should prop just enforce that a null header means the empty tree in HC

PR #586, opened 2024-10-18
Rocksdb copy prologue \o/

mafintosh:
was a doozy but oh boy does it fly now

### DIFF
--- a/lib/copy-prologue.js
+++ b/lib/copy-prologue.js
@@ -140,6 +140,10 @@ async function flushBatch (prologue, src, dst, batch) {
     dst.header.tree = prologueToTree(prologue)
   }
 
+  if (userData.length > 0) {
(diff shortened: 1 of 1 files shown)


lib/copy-prologue.js
```diff
@@ -0,0 +1,225 @@
...
+      writer.putBitfieldPage(pageIndex, pages[pageIndex])
+    }
+
+    const pageBuffer = pages[pageIndex]
+    quickbit.set(pageBuffer, getBitfieldOffset(index), true)
+    writer.putBlock(index, value)
+  }
+
+  for (const { key, value } of userData) {
+    writer.setUserData(key, value)
```
chm-diederichs, line 127: atm we still expect `header.userData` to be populated as well

chm-diederichs (approved):

PR #587, opened 2024-10-18
rocksdb: commit overwrites full state

chm-diederichs:


chm-diederichs:
depends on holepunchto/hypercore-storage#26

mafintosh (approved):

lib/bit-interlude.js
```diff
@@ -38,25 +37,35 @@ module.exports = class BitInterlude {
...
+      if (r.start <= start && end <= r.end) {
+        if (value === r.value) return
+
+        const lower = { start: r.start, end: start, value: r.value }
+        const upper = { start: end, end: r.end, value: r.value }
+
+        let remove = 1
+
+        if (lower.start < lower.end) this.ranges.splice(i++, remove--, lower)
+        this.ranges.splice(i++, remove, { start, end, value })
```
mafintosh, line 64: the ++ is sus to me, we use i again below but normally you'd i-- it as you are removing from the array

lib/bit-interlude.js
```diff
@@ -97,41 +113,49 @@ module.exports = class BitInterlude {
...
+      if (page) {
+        for (let i = 0; i < page.bitfield.length; i++) {
+          view.setUint32(i * 4, page.bitfield[i], true)
+        }
       }
 
-      const last = (page.index + 1) * (buf.byteLength << 3)
-      const offset = page.index * buf.byteLength << 3
+      const last = (pageIndex + 1) * (buf.byteLength << 3)
+      const offset = pageIndex * buf.byteLength << 3
```
mafintosh, line 146: add parens around the buf.. <<  3 part

lib/core.js
```diff
@@ -853,42 +1011,56 @@ module.exports = class Core {
         }
       }
 
-      const promises = []
+      if (this.verifier === null) {
+        throw INVALID_OPERATION('Cannot commit without manifest') // easier to assert than upsert
+      } else if (this.tree.length < length && !signature) {
```
mafintosh, line 1016: remove else, just two ifs

lib/bit-interlude.js
chm-diederichs: only one of them removes from the array, but they all add to it, so it ends up being: 

```
  arr = [0]
  arr.splice(i++, 1, 1)
  arr.splice(i++, 0, 2)
  arr.splice(i, 0, 3)

  // arr === [1, 2, 3]
```

PR #596, opened 2024-11-15
Fix events on rocks

mafintosh:


lib/core.js
```diff
@@ -130,26 +120,27 @@ class Update {
...
     this.name = name
+    this.sessions = []
 
     this.mutex = new Mutex()
 
     this.blocks = blocks
     this.tree = tree
     this.bitfield = bitfield
     this.snapshotLength = snapshotLength
+    this.snapshotCompatLength = snapshotLength
```
chm-diederichs, line 143: do we still need snapshotCompatLength separately?

lib/core.js
```diff
@@ -1302,71 +1352,13 @@ module.exports = class Core {
...
-    }
-  }
-
   async _onconflict (proof, from) {
     await this.replicator.onconflict(from)
 
-    for (const s of this.sessions) s.emit('conflict', proof.upgrade.length, proof.fork, proof)
+    for (let i = this.monitors.length - 1; i >= 0; i--) {
+      const s = this.monitors[i]
+      s.emit('conflict', proof.upgrade.length, proof.fork, proof)
```
chm-diederichs, line 1360: shouldn't this be emitted by all sessions?

lib/core.js
mafintosh: ya cause the normal one is used for the actual length and the compat is how much you can reuse (i think)

lib/core.js
chm-diederichs: ahh sorry i misread the first line of maybeAddMonitor

chm-diederichs (approved):

PR #597, opened 2024-11-16
Add hotswaps

mafintosh:


lib/replicator.js
```diff
@@ -1834,10 +1845,18 @@ module.exports = class Replicator {
     if (b === null) return false
 
     removeInflight(b.inflight, req)
+    removeHotswap(b)
     b.queued = false
 
     b.resolve(value)
 
+    if (b.inflight.length > 0) {
```
HDegroote, line 1853: (detail) The if is not strictly needed here, the for does it implicitly

lib/replicator.js
```diff
@@ -2161,6 +2180,16 @@ module.exports = class Replicator {
     return false
   }
 
+  _updateHotswap (peer) {
+    const maxHotswaps = peer.getMaxHotswapInflight()
+    if (!peer.isActive() || peer.inflight >= maxHotswaps) return
+
+    for (const b of this.hotswaps.pick(peer)) {
+      if (peer._requestBlock(b) === false) continue
+      if (peer.inflight >= maxHotswaps) break
```
HDegroote, line 2189: (sanity check) can ` _requestBlock` never result in additional inflights when it returns `false`? I'm tracing the `_maybeWant` path, and it's not trivial to figure out (`_maybeWant` can send a `wireWant` message, though I don't think that results in an `inflight` increment)

mafintosh: only block requests count as inflight atm, we should prop rename post rocks to blocksInflight for clarity

lib/replicator.js
mafintosh: true, let me add a comment, why not

HDegroote:
Sanity check (quicker than fully tracing it myself): https://github.com/holepunchto/hypercore/blob/69d15e98caa8145e37b0c94c9239567bb2753086/lib/replicator.js#L1973-L1977
=> Do we not need to remove the hotswap here, and/or handle the case where this is the answer to a hotswap request?

mafintosh:
Checking

mafintosh:
We do, nice one

lib/hotswap-queue.js
```diff
@@ -0,0 +1,60 @@
...
+        yield b
+
+        if (--ticks <= 0) break
+      }
+    }
+  }
+
+  add (block) {
+    if (block.hotswap !== null) this.remove(block)
+    if (block.inflight.length === 0 || block.inflight.length >= 3) return
```
HDegroote, line 35: I think `blocks.inflight.length` can never be 0: it's only called from `_sendBlockRequest`, after a req was pushed to the inflights, so I'm not sure what it protects against

mafintosh: just a sanity check cause sep file

mafintosh: cause it makes the queue selection explode, could be an assertion obvs

PR #604, opened 2024-11-28
Expose user data stream from lib/core.js

chm-diederichs:


lib/core.js
```diff
@@ -84,6 +84,11 @@ module.exports = class Core {
     }
   }
 
+  getUserDataStream (opts) {
+    const storage = opts && opts.session ? opts.session.state.storage : this.storage
```
mafintosh, line 88: can you put a paran around the (opts && opts.session) ? always have a hard time understand what wins with these

PR #609, opened 2024-11-29
rocksdb: SessionState.moveTo API

chm-diederichs:


index.js
```diff
@@ -326,6 +328,7 @@ class Hypercore extends EventEmitter {
     }
 
     this.state.addSession(this)
+    this._sessionOpen = true
```
mafintosh, line 331: instead of using this flag you can just check that stateIndex !== -1

index.js
```diff
@@ -299,7 +301,7 @@ class Hypercore extends EventEmitter {
     }
 
     if (opts.parent) {
-      if (opts.parent.state === null) await opts.parent.ready()
+      if (!opts.parent._sessionOpen) await opts.parent.ready()
```
mafintosh, line 304: see comment below but opts.parent._stateIndex === -1 should do the trick

mafintosh (approved):

PR #610, opened 2024-12-02
rocksdb: use storage API for updating dependency

chm-diederichs:


chm-diederichs:
depends on holepunchto/hypercore-storage#40

mafintosh (approved):

PR #611, opened 2024-12-04
rocksdb: always ref state and explicit snapshot lengths

chm-diederichs:


chm-diederichs:
depends on https://github.com/holepunchto/hypercore-storage/pull/41

PR #619, opened 2025-01-10
rocksdb: atomic sessions

chm-diederichs:


index.js
```diff
@@ -453,6 +456,11 @@ class Hypercore extends EventEmitter {
     this.emit('close', true)
   }
 
+  async commit (state, opts) {
```
mafintosh, line 459: should we make this commit(seesion, opts) while we are at it and sniff the state from the session?

PR #621, opened 2025-01-14
v11 features needed by Autobase

chm-diederichs:


lib/session-state.js
```diff
@@ -670,7 +697,7 @@ module.exports = class SessionState {
       await state.mutex.lock()
       srcLocked = true
 
-      await this.core._validateCommit(state, treeLength)
+      if (!await this.core._validateCommit(state, treeLength)) return null
```
mafintosh, line 700: can you put the !await fn() in a paren? !(await fn()) - i find that easier to read

mafintosh (approved):

PR #623, opened 2025-01-20
Add tests on the messages exchanged when peers start replicating a core

HDegroote:


test/replicate.js
```diff
@@ -1688,6 +1690,175 @@ test('uses hotswaps to avoid long download tail', async t => {
...
+  const a = await create(t)
+  await a.append(['a', 'b', 'c', 'd', 'e'])
+  const b = await create(t, a.key)
+
+  replicate(a, b, t)
+  await new Promise(resolve => setTimeout(resolve, 100))
+
+  const msgs = b.replicator.peers[0].stats
+  t.is(msgs.wireSync.tx, 3, 'wire syncs tx')
+  t.is(msgs.wireSync.rx, 3, 'wire syncs rx')
```
HDegroote, line 1703: The non-sparse core might be oversending sync messages. It sends the same sync message 3 times:
```
 {
  fork: 0,
  length: 5,
  remoteLength: 0,
  canUpgrade: true,
  uploading: true,
  downloading: true,
  hasManifest: true
}
```

The empty cores first sends 

```
{
  fork: 0,
  length: 0,
  remoteLength: 0,
  canUpgrade: true,
  uploading: true,
  downloading: true,
  hasManifest: false
}
```
It then sends a second message with updated remoteLength, and a third where both remoteLength and length are set to 5. That seems ok

PR #625, opened 2025-01-21
Test for bitfield messages on cache miss

HDegroote:


test/remote-bitfield.js
```diff
@@ -58,6 +58,70 @@ test('remote congituous length consistency (remote-bitfield findFirst edge case)
...
+  await t.exception(
+    async () => {
+      await empty.get(5, { timeout: 100 })
+    },
+    /REQUEST_TIMEOUT/,
+    'request on unavailable block times out (sanity check)'
+  )
+
+  // TODO: verify this is expected
+  t.is(stats.wireBitfield.rx, 1, 'No new bitfield message sent on new cache miss')
```
HDegroote, line 121: Is this expected? I had expected a new bitfield message, where the sparse peer communicates which blocks it now has available (both peers did send a new wireSync message to each other, so they're aware of each other's lengths)

HDegroote: confirmed it's expected, so just got rid of all those additional asserts

PR #628, opened 2025-01-22
defer closing of moved storage to end to avoid read races

mafintosh:


lib/session-state.js
```diff
@@ -244,7 +252,8 @@ module.exports = class SessionState {
 
       this._moveToCore(src.core)
 
-      await oldStorage.close()
+      if (this.lingers === null) this.lingers = []
+      this.lingers.push(oldStorage)
```
chm-diederichs, line 256: we could add to lingers at the start of the scope and avoid assigning oldStorage

mafintosh: nice

chm-diederichs (approved):

PR #632, opened 2025-02-04
copy existing bits when overwriting

chm-diederichs:


chm-diederichs:
depends on https://github.com/holepunchto/hypercore-storage/pull/68

PR #635, opened 2025-02-10
remove dependencyLength property

chm-diederichs:


chm-diederichs:
depends on https://github.com/holepunchto/hypercore-storage/pull/71

PR #640, opened 2025-02-16
Public unencrypted flag

mafintosh:


mafintosh:
requires an update to hypercore-storage as well to store the new flag

index.js
```diff
@@ -230,6 +230,7 @@ class Hypercore extends EventEmitter {
 
   async setEncryptionKey (encryptionKey, opts) {
     if (!this.opened) await this.opening
+    if (this.core.unencrypted) return
```
chm-diederichs, line 233: should we throw an error?

index.js
```diff
@@ -759,7 +760,8 @@ class Hypercore extends EventEmitter {
       block = b4a.from(block)
 
       if (this.encryption.compat !== this.core.compat) this._updateEncryption()
-      this.encryption.decrypt(index, block)
+      if (this.core.unencrypted) this.encryption = null
```
chm-diederichs, line 763: if we're storing in the manifest then it shouldn't be possible to dynamically set unencrypted right?

chm-diederichs: ahh sorry, this is in the case that we do:
```
const core = new Hypercore(unencryptedManifest)
const sess = core.session({ encryptionKey }
```

in which case, should we throw an error?

index.js
mafintosh: easier to ignore it for autobase atm, but post some engineering, thats prop the way yes

index.js
mafintosh: ya its never dynamically set, we check it here for replication reasons as the manifest might arrive post-ready

PR #650, opened 2025-02-26
Support checkouts for named sessions only

HDegroote:
Currently it's unsafe to use the checkout option without the name option. This PR makes it explicit that this is not supported.


index.js
```diff
@@ -208,6 +208,9 @@ class Hypercore extends EventEmitter {
       // in practice, open an issue and we'll try to make a solution for it.
       throw SESSION_CLOSED('Cannot make sessions on a closing core')
     }
+    if (opts.checkout && !opts.name) {
```
mafintosh, line 211: checkout !== undefined

HDegroote: Fixed (now does explicit undefined checks for the others too)

PR #651, opened 2025-02-26
Fix flaky 'remote has larger tree' test

HDegroote:
The test was flaky in a subtle way:

- The intent of `c.get(4) // unresolvable but should not crash anything` was to get a block that's not resolvable
- But that block was in fact resolvable (fixed to `c.get(5)` in this PR)
- This caused a low-likelihood race condition, which is the actual bug: `core.get(i)` should always have a catch handler, because any pending requests are cancelled with an error when the hypercore closes. The error triggered when the request had not yet resolved by the time test teardown started.

Fixed by requesting the correct block, and adding a catch handler

### DIFF
--- a/test/helpers/index.js
+++ b/test/helpers/index.js
@@ -13,6 +13,7 @@ exports.create = async function (t, ...args) {
   await core.ready()
 
   t.teardown(() => core.close(), { order: 1 })
+  // t.teardown(() => core.close().catch(safetyCatch), { order: 1 })
 
   return core
 }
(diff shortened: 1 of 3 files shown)


test/helpers/index.js
```diff
@@ -12,7 +12,7 @@ exports.create = async function (t, ...args) {
   const core = new Hypercore(db, ...args)
   await core.ready()
 
-  t.teardown(() => core.close(), { order: 1 })
+  t.teardown(() => core.close().catch(safetyCatch), { order: 1 })
```
HDegroote, line 15: Note: this is not related to this PR, and I can remove it. But I think we should either
- Catch the error (assuming we allow `core.close()` to throw)
- Fail the test if `core.close()` throws (assuming it is not allowed to throw)

Went for catching, since I think it's allowed to throw

PR #653, opened 2025-02-27
overwrite should always write roots

chm-diederichs:


lib/session-state.js
```diff
@@ -754,6 +754,11 @@ module.exports = class SessionState {
       tx.deleteBlockRange(treeLength, this.length)
     }
 
+    for (const root of roots) {
+      if (root !== null) tx.putTreeNode(root)
```
mafintosh, line 758: a root can never but null as we use them to gen the hash below

PR #655, opened 2025-02-27
Update persisted contiguousLength on append

HDegroote:
The previous flow is to set the persisted contiguousLength when the hypercore next loads, in `Hypercore.core._tryOpen()`. This updates it when it is actually set.


mafintosh:
We should refactor so we set it whenever we set the bitfield, leaving open until then (good pr tho)

chm-diederichs (approved):

PR #661, opened 2025-03-03
Safe lock usage for reorg

HDegroote:
In other functions, `createWriteBatch` lives inside the try too. This avoids deadlock should `createWriteBatch` ever throw


HDegroote:
Unneeded change, since it is an invariant that it doesn't throw post mutex acquisition

PR #675, opened 2025-03-12
Do not include bool when emitting a close event

HDegroote:
Technically a breaking change, but this boolean argument was not documented.

Removing because it is no longer correctly indicating whether this was the last session (due to the concept of weak sessions, like in the passive-core-watcher)


mafintosh:
nice one

PR #676, opened 2025-03-12
Use async `hypercore-encryption` module

chm-diederichs:


chm-diederichs:
we need to publish https://github.com/holepunchto/hypercore-block-encryption once we're happy with it

index.js
```diff
@@ -1119,3 +1140,16 @@ function getEncryptionOption (opts) {
   if (!opts.encryption) return null
   return b4a.isBuffer(opts.encryption) ? { key: opts.encryption } : opts.encryption
 }
+
+function isEncryptionProvider (e) {
+  return !!(e && (e instanceof HypercoreEncryption || (e.encrypt && typeof e.encrypt === 'function')))
```
mafintosh, line 1145: you can just do the instanceof check

index.js
```diff
@@ -1046,14 +1077,14 @@ function toHex (buf) {
...
 }
 
-function preappend (blocks) {
+async function preappend (blocks) {
   const offset = this.state.length
   const fork = this.state.encryptionFork
 
   if (this.encryption.compat !== this.core.compat) this._updateEncryption()
 
   for (let i = 0; i < blocks.length; i++) {
```
RangerMauve, line 1086: Should this maybe be done in parallel using something like `await Promise.all(blocks.map(encrypt))`? This seems like a major perf regression otherwise.

mafintosh: same as today since sync right now, but we can adjust later. the async'ness is there for flexibility. for now lets keep it simple.

PR #680, opened 2025-03-26
unref all timers

mafintosh:


mafintosh:
we went in diff direction

PR #685, opened 2025-04-03
add block request time

MKPLKN:


lib/replicator.js
```diff
@@ -2097,9 +2098,10 @@ module.exports = class Replicator {
   }
 
   _ondata (peer, req, data) {
+    req.elapsed = req.timestamp ? Date.now() - req.timestamp : 0
     if (data.block !== null) {
       this._resolveBlockRequest(this._blocks, data.block.index, data.block.value, req)
-      this._ondownload(data.block.index, data.block.value.byteLength, peer)
+      this._ondownload(data.block.index, data.block.value.byteLength, peer, { elapsed: req.elapsed })
```
mafintosh, line 2104: should we just fwd the req here instead?

lib/replicator.js
```diff
@@ -2097,9 +2098,10 @@ module.exports = class Replicator {
   }
 
   _ondata (peer, req, data) {
+    req.elapsed = req.timestamp ? Date.now() - req.timestamp : 0
```
mafintosh, line 2101: init elapsed and timestamp to 0 where the req is made also for good meassure

lib/replicator.js
MKPLKN: yeah, no reason why not fwd the req instead. Updated

PR #686, opened 2025-04-04
set context on encryption provider

chm-diederichs:


chm-diederichs:
depends on https://github.com/holepunchto/hypercore-encryption/pull/2

lib/core.js
```diff
@@ -114,6 +115,13 @@ module.exports = class Core {
     if (this.exclusive !== null) this.exclusive.unlock()
   }
 
+  _ensureEncryptionContext () {
+    if (this.encryptionContext) return
+    this.encryptionContext = {}
```
mafintosh, line 120: lets ensure the shape here immediately, like so

```
this.encryptionContext = {
  key: this.key,
  manifest: this.manifest
}
```

both are inited to null here so should be fine

index.js
```diff
@@ -1053,22 +1057,51 @@ class Hypercore extends EventEmitter {
...
     const blockKey = block
       ? encryptionKey
       : getLegacyBlockKey(this.key, encryptionKey, this.core.compat)
 
-    return HypercoreEncryption.createLegacyProvider(blockKey)
+    this.core.encryptionContext.blockKey = blockKey
+
+    return new HypercoreEncryption(crypto.hash(blockKey), {
+      getBlockKey: this._getBlockKey.bind(this) // only set once per session
+    })
```
mafintosh, line 1094: shouldnt we make a proper class for this so the block key can just be passed in the constructor and we dont need the runtime bind here?

index.js
```diff
@@ -55,6 +55,7 @@ class Hypercore extends EventEmitter {
 
     this.core = null
     this.state = null
+    this.padding = 0
```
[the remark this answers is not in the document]
chm-diederichs: yes, how come?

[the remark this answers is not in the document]
mafintosh: all good, missed that we are setting it below

index.js
```diff
@@ -1052,23 +1068,33 @@ class Hypercore extends EventEmitter {
...
+
   _updateEncryption () {
     const e = this.encryption
-    if (HypercoreEncryption.isHypercoreEncryption(e)) return
+    if (!HypercoreEncryption.isHypercoreEncryption(e)) {
+      this.encryption = this._getEncryptionProvider(e.key, e.block)
+    }
 
-    this.encryption = this._getLegacyEncryption(e.key, e.block)
+    if (this.padding === 0 || !this.key || !this.manifest) {
```
mafintosh, line 1083: should this be `&& this.key && this.manifest` ?

index.js
```diff
@@ -1052,23 +1079,26 @@ class Hypercore extends EventEmitter {
...
-    const e = this.encryption
-    if (HypercoreEncryption.isHypercoreEncryption(e)) return
-
-    this.encryption = this._getLegacyEncryption(e.key, e.block)
+  _ensureEncryption () {
+    if (!this.encryption) return
+    this._updatePadding()
 
-    if (e === this.core.encryption) this.core.encryption = this.encryption
+    if (this.encryption.version === -1) return this._loadEncryption()
```
mafintosh, line 1086: just inline loadEncryption?

mafintosh (approved):

PR #689, opened 2025-04-09
Accept user encryption modules and add userData field to v2 manifest

chm-diederichs:


index.js
```diff
@@ -1102,7 +1070,7 @@ class Hypercore extends EventEmitter {
 
   _getEncryptionProvider (encryptionKey, block) {
     if (!encryptionKey) return null
-    return new DefaultEncryption(encryptionKey, block)
+    return new DefaultEncryption(encryptionKey, this.key, { block, compat: this.core.compat })
```
mafintosh, line 1073: is this called just in time? ie just before decrypt? cause it only knows the compat flag post replication

lib/default-encryption.js
```diff
@@ -0,0 +1,108 @@
+const sodium = require('sodium-universal')
+const c = require('compact-encoding')
+const b4a = require('b4a')
+const { LEGACY_BLOCK_ENCRYPTION } = require('./caps')
+
+const nonce = b4a.alloc(sodium.crypto_stream_NONCEBYTES)
+const PADDING = 8
+
+module.exports = class DefaultEncryption {
+  static padding = 8
```
mafintosh, line 10: nit: padding = PADDING

mafintosh (approved):

PR #697, opened 2025-05-12
Win end when live is added in readstream

supersuryaansh:
Currently in `Hypercore.createReadStream({live: true, start: startIndex, end: startIndex }) `  "live" and "end" are two conflicting options, having "end" blocks won't work if live is true.

This PR makes sure that end wins in live vs end.

### DIFF
--- a/test/streams.js
+++ b/test/streams.js
@@ -82,9 +82,6 @@ test('read stream with end and live (live should be ignored)', async function (t
   }
 
   t.alike(collected, expected)
-
-  t.is(collected.includes('delta'), false)
-  t.is(collected.includes('epsilon'), false)
 })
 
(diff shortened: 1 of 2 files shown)


test/streams.js
```diff
@@ -55,6 +55,34 @@ test('read stream with start / end', async function (t) {
...
+
+  for await (const data of stream) {
+    collected.push(b4a.toString(data))
+  }
+
+  t.alike(collected, expected)
+
+  await core.append(['delta', 'epsilon'])
+
+  t.is(collected.includes('delta'), false)
```
mafintosh, line 82: you only push to the stream above so this would never be false, even if end: 3 wasnt set.

PR #704, opened 2025-06-03
Support length opt for update

HDegroote:


HDegroote:
Drafted because we can't get this feature in as-is: we can't use events for control flows, because it will trigger max listener warnings.

PR #709, opened 2025-07-18
WIP: Doc review

lejeunerenard:
`Hypercore` constructor options skipped:
- `weak` : Used mostly in sessions and is already documented there
- `snapshot` : Should be set via `core.snapshot()` instead of manually set.
- `parent` : Used for session management so general not manually set.
- `core` : Likely not manually set except for advanced usecases
- `exclusive` : For the same reasons as `weak`

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,3 +539,56 @@ Emitted when a block is uploaded to a peer.
 #### `core.on('download', index, byteLength, peer)`
 
 Emitted when a block is downloaded from a peer.
+
+#### `Hypercore.MAX_SUGGESTED_BLOCK_SIZE`
+
+The constant for max size (15MB) for blocks appended to Hypercore. This max ensures blocks are replicated smoothly.
+
+#### `const key = Hypercore.key(manifest, options = {})`
+
+Returns the key for a given manifest.
+
+`options` include:
+
+```
+{
+  compat: false,  // Whether the manifest has a single singer whos public key is the key
+  version,        // Manifest version if the manifest argument is the public key of a single singer
+  namespace       // The signer namespace if the manifest argument is the public key of a single singer
+}
+```
+
+#### `const dKey = Hypercore.discoveryKey(key)`
+
+Returns the discovery key for the provided `key`.
+
+#### `const bkey = Hypercore.blockEncryptionKey(key, encryptionKey)`
+
+Returns a block encryption key derived from the `key` and `encryptionKey`.
+
+#### `const mux = Hypercore.getProtocolMuxer(stream)`
+
+Returns a protomux instance from the provided `stream` Hypercore protocol stream.
+
+#### `const core = Hypercore.createCore(storage, opts)`
+
+Returns the internal core using the `storage` and `opts` without creating a full Hypercore instance.
+
+#### `const stream = Hypercore.createProtocolStream(isInitiator, opts = {})`
+
+Create an encrypted noise stream with a protomux instance attached used for Hypercore's replication protocol.
+
+`isInitiator` can be a framed stream, a protomux or a boolean for whether the stream should be the initiator in the noise handshake.
+
+`opts` can include:
+
+```
+{
+  keepAlive: true, // Whether to keep the stream alive
+  ondiscoverykey: () => {}, // A handler for when a discovery key is set over the stream for corestore management
+}
+```
+
+#### `const storage = Hypercore.defaultStorage(storage, opts = {})`
+
+Returns a default hypercore storage. The `storage` argument can be a path where to create the `hypercore-storage` instance or an existing `hypercore-storage` instance. If an existing instance, it is immediately returned.
--- a/README.md
+++ b/README.md
@@ -62,7 +62,9 @@ Alternatively you can pass a [Hypercore Storage](https://github.com/holepunchto/
   notDownloadingLinger: 20000, // How many milliseconds to wait after downloading finishes keeping the connection open. Defaults to a random number between 20-40s
   allowFork: true, // Enables updating core when it forks
   manifest: undefined, // Set the manifest when creating the hypercore. See Manifest section for more info
-  preload: undefined // A promise that returns constructor options overrides before the core is opened
+  preload: undefined, // A promise that returns constructor options overrides before the core is opened
+  storage: undefined, // An alternative to passing storage as a dedicated argument
+  key: null, // An alternative to passing key as a dedicated argument
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -61,7 +61,8 @@ Alternatively you can pass a [Hypercore Storage](https://github.com/holepunchto/
   ongc: (session) => { ... }, // A callback called when the session is garbage collected
   notDownloadingLinger: 20000, // How many milliseconds to wait after downloading finishes keeping the connection open. Defaults to a random number between 20-40s
   allowFork: true, // Enables updating core when it forks
-  manifest: undefined // Set the manifest when creating the hypercore. See Manifest section for more info
+  manifest: undefined, // Set the manifest when creating the hypercore. See Manifest section for more info
+  preload: undefined // A promise that returns constructor options overrides before the core is opened
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ Alternatively you can pass a [Hypercore Storage](https://github.com/holepunchto/
   ongc: (session) => { ... }, // A callback called when the session is garbage collected
   notDownloadingLinger: 20000, // How many milliseconds to wait after downloading finishes keeping the connection open. Defaults to a random number between 20-40s
   allowFork: true, // Enables updating core when it forks
+  manifest: undefined // Set the manifest when creating the hypercore. See Manifest section for more info
 }
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,9 +60,14 @@ Alternatively you can pass a [Hypercore Storage](https://github.com/holepunchto/
   writable: true, // disable appends and truncates
   inflightRange: null, // Advanced option. Set to [minInflight, maxInflight] to change the min and max inflight blocks per peer when downloading.
   ongc: (session) => { ... }, // A callback called when the session is garbage collected
+  onseq: (index, core) => { ... }, // A callback called when core.get(index) is called.
   notDownloadingLinger: 20000, // How many milliseconds to wait after downloading finishes keeping the connection open. Defaults to a random number between 20-40s
   allowFork: true, // Enables updating core when it forks
   userData: { foo: 'bar' }, // An object to assign to the local User Uata storage
+  manifest: undefined, // Set the manifest when creating the hypercore. See Manifest section for more info
+  preload: undefined, // A promise that returns constructor options overrides before the core is opened
```
HDegroote: (detail) would mark this as advanced (it's really only there for corestore, normal users won't use this)

README.md
```diff
@@ -79,7 +85,34 @@ Keys are always strings and values can be strings or buffers.
...
+
+```
+{
+  version: 1,                       // Version of the manifest format
+  hash: 'blake2b',                  // Only Blake2b is supported currently
+  allowPatch: false,                // Whether the hypercore can be "patched" to change the signers
+  quorum: (signers.length / 2) + 1, // How many signers needed to verify a block
+  signers,                          // Array of signers for the core
+  prologue: null,                   // The tree hash and length of the core
+  linked: null,                     // Only supported in versions >= 2
```
HDegroote: (detail) Could use an explanation of what `linked` is, likewise for `userData`

README.md
```diff
@@ -79,7 +85,34 @@ Keys are always strings and values can be strings or buffers.
...
+  userData: null                    // Only supported in versions >= 2
+}
+```
+
+Signers are an array of object with the following structure:
```
HDegroote: typo: array of objects

README.md
```diff
@@ -115,8 +159,10 @@ const blockLocal = await core.get(44, { wait: false })
   wait: true, // wait for block to be downloaded
   onwait: () => {}, // hook that is called if the get is waiting for download
   timeout: 0, // wait at max some milliseconds (0 means no timeout)
+  activeRequests: undefined, // Advanced option. Pass BlockRequest for replicating the block
```
HDegroote, line 164: Question, not a comment: do you know what this is used for?

README.md
```diff
@@ -138,7 +184,9 @@ console.log('core was updated?', updated, 'length is', core.length)
 
 ``` js
 {
-  wait: false
+  wait: false,
+  activeRequests: undefined, // Advanced option. Pass requests for replicating blocks
+  force: false, // Force an update even if core is writable.
```
HDegroote, line 191: Typo: 'even if core is not writable'

README.md
```diff
@@ -220,12 +271,29 @@ partialStream.pipe(process.stdout)
...
 ```
 
+#### `const stream = core.createWriteStream()`
+
+Make a write stream to append chunks as blocks.
+
+``` js
+const ws = core.createWriteStream()
+
+for (const data of ['hello', 'world']) ws.write(data)
```
HDegroote, line 292: I'm not sure this usage is safe. If there is an error (for example because the storage is out of space), won't this throw an unhandled rejection?

README.md
```diff
@@ -220,12 +271,29 @@ partialStream.pipe(process.stdout)
...
+ws.end()
+
+// Wait for stream to finish
+await new Promise(resolve => ws.on('finish', resolve))
```
HDegroote: I'm not sure this sequence is safe:
```
ws.end()

// Wait for stream to finish
await new Promise(resolve => ws.on('finish', resolve))
```
I didn't trace fully in streamx, so if you're sure then it's fine to leave. But if not: better to start listening on the 'finish' before calling `ws.end()`

README.md
```diff
@@ -244,17 +312,55 @@ The core will also gossip to peers it is connected to, that is no longer has the
...
 
+#### `const proof = await core.proof(opts)`
+
+Generate a proof (a `TreeProof` instance) for the request `opts`.
+
+`opts` include:
+
+```
+{
+  block: { index: 4, nodes: 2 }, // Block request
```
HDegroote: These numbers are example numbers, not default numbersm right? I would just specify it as `block: { index , nodes },`, maybe clarifying that they're both integers

(elsewhere in the readme we use this notation for specifying default values)

README.md
```diff
@@ -244,17 +312,55 @@ The core will also gossip to peers it is connected to, that is no longer has the
...
+  seek: { bytes: 13, padding: 0 }, // Seek Request
+  upgrade: { start: 8, length: 2 } // Upgrade request
+}
+```
+
+Note that you cannot seek & provide a block / hash request when upgrading.
+
+#### `const batch = await core.verifyFullyRemote(proof)`
+
+Return merkle tree batch from proof. Will throw if proof cannot be verified.
```
HDegroote: (detail) missing articles `Return the merkle tree batch from the proof. Will throw if the proof cannot be verified.`

README.md
```diff
@@ -342,9 +474,20 @@ Attempt to apply blocks from the session to the `core`. `core` must be a default
 
 Returns `null` if committing failed.
 
+`opts` includes:
+
+```
+{
+  length: session.length, // the core's length after committing the blocks
```
HDegroote, line 485: Should these `length` and `treeLength` options ever be overwritten? I don't know this code, but my first guess is that everything would break

README.md
```diff
@@ -464,6 +607,34 @@ Populated after `ready` has been emitted. Will be `0` before the event.
...
+
+```
+{
+  block: false, // Whether the key is for block encryption
+}
+```
+
+#### `core.setKeyPair(keyPair)`
+
+Update the core's `keyPair`.
```
HDegroote: Is this a safe operation, or should we add some caveats here?

README.md
```diff
@@ -464,6 +607,34 @@ Populated after `ready` has been emitted. Will be `0` before the event.
...
+
+#### `core.setActive(active)`
+
+Set the core to be active or not. A core is considered 'active' if it should linger to download blocks from peers.
```
HDegroote, line 640: Note that when users call `setActive(true), they have to call `setActive(false)` when they're done, otherwise the book keeping breaks.

README.md
```diff
@@ -531,3 +704,56 @@ Emitted when a block is uploaded to a peer.
...
+
+#### `const key = Hypercore.key(manifest, options = {})`
+
+Returns the key for a given manifest.
+
+`options` include:
+
+```
+{
+  compat: false,  // Whether the manifest has a single singer whos public key is the key
```
HDegroote: typo: whose

README.md
lejeunerenard: I believe this is not a typo. Looking at this line, https://github.com/holepunchto/hypercore/blob/252c7c14c9d57e2baf9c5b541987f6f916b5331c/index.js#L677
it only applies if the the cores is writable.

I believe this is because a writable core shouldn't need to update itself since all appends are assumed to be local.

README.md
lejeunerenard: Good point. Marked them as advanced.

README.md
lejeunerenard: Done!

README.md
lejeunerenard: Fixed thanks!

README.md
lejeunerenard: This can be used in `hyperdb` here:
https://github.com/holepunchto/hyperdb/blob/c2039a11b272e3527a8ada177e569b20b592fd58/lib/engine/bee.js#L220

`hyperbee` calls `core.get(seq, { activeRequests })` here:
https://github.com/holepunchto/hyperbee/blob/c907231f1bfe517cf18c7294a605c350a625d8d6/index.js#L921

README.md
lejeunerenard: I think in this case, the example is fine without protection against side effects / errors. For simplicity the example can be a 'happy path' version where everything is assumed to work. It just can't teach bad patterns. Error handling will depend on the use case so doesn't count as a bad pattern IMHO.

For reference, Node.js docs often exclude try/catch for the same type of examples:
https://nodejs.org/api/stream.html#writableendchunk-encoding-callback

README.md
lejeunerenard: It works in this scenario because all writes/appends are async and the `finish` event doesn't fire on calling `ws.end()` since it waits for the buffer to flush.

I think though that you're right that the `finish` should be before earlier since someone might implement it naively and end up adding something that does making waiting for the `finish` unsafe. This could be considered a bad pattern (as previously mentioned) even though it works here. I've adjusted accordingly.

README.md
lejeunerenard: Yeah you're right. I added them to convey the types of the nested properties, but that would be inconsistent. I really should have a consistent style guide for api docs.
Fixed thanks!

README.md
lejeunerenard: Fixed thanks!

README.md
lejeunerenard: Yeah for `length` you can partially flush a commit. See this test case:
https://github.com/holepunchto/hypercore/blob/252c7c14c9d57e2baf9c5b541987f6f916b5331c/test/batch.js#L234-L260
For `treeLength` there is this test case for using it with atoms:
https://github.com/holepunchto/hypercore/blob/252c7c14c9d57e2baf9c5b541987f6f916b5331c/test/atomic.js#L197-L245

There is only one place I can find that uses either outside of tests and thats in autobase:
https://github.com/holepunchto/autobase/blob/f0f0a687adb4341e9390d1dc14c0deb9fc60df4e/lib/store.js#L137
But this is the store reimplementing the commit method for a view core which is used when flushing apply state here:

https://github.com/holepunchto/autobase/blob/f0f0a687adb4341e9390d1dc14c0deb9fc60df4e/lib/apply-state.js#L923

> ... my first guess is that everything would break

There are safe guards in place ( see `commit()` in `session-state.js` [here](https://github.com/holepunchto/hypercore/blob/252c7c14c9d57e2baf9c5b541987f6f916b5331c/lib/session-state.js#L835) ), but if you can think of a way to break it, then we should protect against that.

README.md
lejeunerenard: Great question. Definitely should be marked as advanced. Kinda assumed that anyone messing with key pair would treat it with white gloves since all key pairs should which is not a safe assumption.

Clarified and added a test case to demonstrate at least one footgun.

README.md
lejeunerenard: Very true. Thanks! Added.

README.md
lejeunerenard: Thanks fixed!

lejeunerenard:
@HDegroote thank you for a thorough review! Finding all the spelling mistakes and potential pitfalls was very helpful!

README.md
HDegroote: Ah yes, my bad, sorry!

README.md
HDegroote: Nice, thanks for clarifying!

test/basic.js
```diff
@@ -763,6 +764,20 @@ test('append alignment to bitfield boundary', async function (t) {
   }
 })
 
+test('setKeyPair', async function (t) {
```
HDegroote: (detail) I'd remove this test, since it's not really related to this PR and it's already a very big PR (fine to PR separately though, particularly if we don't have tests for setKeyPair already)

HDegroote (approved):
Looks good to me

test/basic.js
lejeunerenard: Removed

PR #712, opened 2025-08-02
package-lock.json: include in repository

ethancedwards8:
It is generally advised to include the package-lock.json files in repositories so ensure the correct dependencies are installed.

Additionally, in Nixpkgs it is helpful for us to have lockfiles included upstream rather than us having to create one and vendor it in Nixpkgs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 node_modules
-package-lock.json
 .DS_Store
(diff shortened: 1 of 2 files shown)


mafintosh:
Thanks but we only use lock files in apps never modules

PR #715, opened 2025-08-16
allow a bit more flexibility in request handing

mafintosh:


lib/messages.js
```diff
@@ -514,13 +514,20 @@ wire.data = {
 wire.noData = {
   preencode (state, m) {
     c.uint.preencode(state, m.request)
+    c.uint.preencode(state, m.reason ? 1 : 0)
```
HDegroote, line 517: I think this is fine, but making sure: this is future-proof in the sense that we can add more flags later,, correct? (by adding 2, 4, 8, 16 etc))

mafintosh: exactly, also hyperschema compat which we should move this to in a separate pr

PR #718, opened 2025-08-20
simplify the requestable condition

mafintosh:


lib/replicator.js
```diff
@@ -1233,42 +1235,10 @@ class Peer {
     return false
   }
 
-  _hasTreeParent (index) {
+  _canRequest (index) {
```
HDegroote, line 1242: What do you think of leaving the assertion `if (!(index >= 0)) throw ASSERTION('bad index to _hasTreeParent: ' + index)`, in case a negative index was indeed the cause of the issue? Because if the index is ever negative, it indicates a bug in some other part of the code.

lib/replicator.js
```diff
@@ -662,6 +663,7 @@ class Peer {
 
     if (length === this.remoteLength && fork === this.core.state.fork) {
       this.canUpgrade = upgrade
+      if (upgrade) this.lastUpgradableLength = length
```
HDegroote: I think we also need logic for the case where the hypercore gets forked, in which case the `lastUpgradableLength` of the previous fork loses meaning.

HDegroote: Also verifying: `lastUpgradableLength` will be reduced when there is a truncation, right?

lib/replicator.js
mafintosh: true

lib/replicator.js
mafintosh: also true

mafintosh: mb easiest is just storing lastUpgradableFork also and check both in canRequest?

lib/replicator.js
HDegroote: Done

lib/replicator.js
HDegroote: Added

PR #723, opened 2025-08-27
Broadcast range to all peers when a core becomes fully available

HDegroote:
This is a more efficient version of https://github.com/holepunchto/hypercore/pull/722, because we now send the broadcast only once.

It does require more book keeping.

### DIFF
--- a/test/replicate.js
+++ b/test/replicate.js
@@ -2155,13 +2155,13 @@ test('range is broadcast when a core is fully available', async function (t) {
   await writer.append(['f', 'g'])
   // Give time for messages to be received
   await new Promise(resolve => setTimeout(resolve, 100))
-  t.is(reader.replicator.peers[0]._remoteContiguousLength, 1, 'reader detected writer no longer fully contig')
+  t.is(reader.replicator.peers[0].remoteContiguousLength, 1, 'reader detected writer no longer fully contig')
 
   await reader.get(5)
   await reader.get(6)
   // Give time for broadcastRange message to be received
   await new Promise(resolve => setTimeout(resolve, 100))
-  t.is(writer.replicator.peers[0]._remoteContiguousLength, 7, 'writer detected reader is fully contig')
+  t.is(writer.replicator.peers[0].remoteContiguousLength, 7, 'writer detected reader is fully contig')
 })
 
 test('range is broadcast when a core is fully available (multiple peers)', async function (t) {
@@ -2192,10 +2192,10 @@ test('range is broadcast when a core is fully available (multiple peers)', async
 
   await new Promise(resolve => setTimeout(resolve, 100))
 
-  t.is(writerToReader1._remoteContiguousLength, 0, 'reader1 skipped range update to writer (not contig yet)')
-  t.is(writerToReader2._remoteContiguousLength, 0, 'reader2 skipped range update to writer (not contig yet)')
-  t.is(reader1ToWriter._remoteContiguousLength, 3, 'writer updated for reader1')
-  t.is(reader2ToWriter._remoteContiguousLength, 3, 'writer updated for reader2')
+  t.is(writerToReader1.remoteContiguousLength, 0, 'reader1 skipped range update to writer (not contig yet)')
+  t.is(writerToReader2.remoteContiguousLength, 0, 'reader2 skipped range update to writer (not contig yet)')
+  t.is(reader1ToWriter.remoteContiguousLength, 3, 'writer updated for reader1')
+  t.is(reader2ToWriter.remoteContiguousLength, 3, 'writer updated for reader2')
 
   await reader1.get(1)
   await reader1.get(2)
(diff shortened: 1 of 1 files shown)


lib/replicator.js
```diff
@@ -2156,7 +2157,15 @@ module.exports = class Replicator {
...
   }
 
+  get hasFullCoreLocally () {
+    if (!this.core.state) return false
+    if (!this.core?.header?.hints?.contiguousLength) return false
+    return this.core.state.length === this.core.header.hints.contiguousLength
+  }
+
   _ondata (peer, req, data) {
+    if (!this.hasFullCoreLocally && this.fullyDownloaded) this.fullyDownloaded = false
```
HDegroote: I'm not very happy with invalidating `this.fullyDownloaded` here. The clean approach would be to detect when the core's length updates, but afaik there's no hook for that.

Doing it here works because it's the only place we need it, but it's not future proof, so open to ideas.

test/remote-length.js
```diff
@@ -29,10 +29,10 @@ test('contiguous-length announce-on-update flow', async function (t) {
   replicate(a, b, t)
   replicate(b, c, t)
 
-  await a.append('a')
+  await a.append(['a', 'b'])
```
HDegroote, line 32: This test verifies that we do not gossip when we know the remote already has that range. This PR adds an exception to that rule, so I had to modify the test to not hit that exception (by making sure the reader core `b` does not have all blocks)

lib/replicator.js
```diff
@@ -1605,7 +1606,24 @@ module.exports = class Replicator {
 
   // Called externally when a range of new blocks has been processed/removed
   onhave (start, length, drop = false) {
-    for (const peer of this.peers) peer.broadcastRange(start, length, drop)
+    let force = false
+    if (!drop) {
+      const contig = this.core?.header?.hints?.contiguousLength
+
+      // Send the biggest range we can, rather than the incremental upgrade
```
HDegroote: Before when we're appending block-per-block to a core, it would send range messages like
- 5, 6, false
- 6, 7, false
- 7, 8, false
- ...

Now it sends:
- 0, 6, false
- 0, 7, false
- 0, 8, false
- ...

HDegroote: Note: we were already doing that optimisation in `broadcastRange` itself. I'm checking if we can't put all this logic there

lib/replicator.js
HDegroote: Found a cleaner approach

lib/replicator.js
HDegroote: All logic is now on broadcastRange, so this thread is now irrelevant

lib/replicator.js
```diff
@@ -2156,6 +2165,12 @@ module.exports = class Replicator {
     for (const peer of this.peers) peer._resetMissingBlock(index)
   }
 
+  get hasFullCore () {
+    if (!this.core.state) return false
+    if (!this.core?.header?.hints?.contiguousLength) return false
```
mafintosh: no need for ? - all are set always post open

lib/replicator.js
```diff
@@ -2156,6 +2165,12 @@ module.exports = class Replicator {
     for (const peer of this.peers) peer._resetMissingBlock(index)
   }
 
+  get hasFullCore () {
```
mafintosh: fullyDownloaded () { is more in line with the rest of your naming imo and make it a non getter

lib/replicator.js
```diff
@@ -482,6 +483,7 @@ class Peer {
   }
 
   signalUpgrade () {
+    if (!this.replicator.hasFullCore && this.fullyDownloaded) this.fullyDownloaded = false
```
mafintosh: nit: i would check the bool first in the if and then call the replicator method, less overhead

test/replicate.js
```diff
@@ -2122,6 +2122,100 @@ test('download event includes "elapsed" time in metadata', async function (t) {
...
+  await new Promise(resolve => setTimeout(resolve, 100))
+
+  const reader1ToReader2 = reader1.replicator.peers[1]
+  const reader2ToReader1 = reader2.replicator.peers[1]
+  t.is(reader1ToReader2 !== undefined, true, 'sanity check')
+  t.is(reader2ToReader1 !== undefined, true, 'sanity check')
+
+  await new Promise(resolve => setTimeout(resolve, 100))
+
+  t.is(writerToReader1._remoteContiguousLength, 0, 'reader1 skipped range update to writer (not contig yet)')
```
mafintosh: for these why do we need to check the _ prop? we dont have a public version of this?

lib/replicator.js
HDegroote: Fixed (means I could simplify the method too)

lib/replicator.js
HDegroote: done, and renamed the one on Peer to `fullyDownloadedSignaled`

test/replicate.js
HDegroote: We do, fixed

PR #731, opened 2025-09-17
Avoid uncaught promises by awaiting multiple in objs

lejeunerenard:
The following illustrates how this can occur:

```js
async function test () {
  const prom1 = new Promise((resolve) => setTimeout(resolve, 100))
  const prom2 = Promise.reject(Error('whoops'))

  return {
    a: await prom1,
    b: await prom2
  }
}

test().catch(() => console.log('catch'))
```

The `whoops` will reject before `prom1` is awaited causing it to be uncaught.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -815,9 +815,10 @@ async function getCoreInfo (storage) {
 
   r.tryFlush()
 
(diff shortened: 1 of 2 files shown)


mafintosh:
This is also safe as is

PR #736, opened 2025-09-25
WIP: Refactor bitfield to replicator & reduce bitfield memory by relying on storage

lejeunerenard:
This PR currently is only the first step of moving the bitfield into the replicator. Since the bitfield is only required by the replicator for checking local vs remote blocks.

This refactor will require updating the following known locations:

- [`Hyperbee`'s `preload`](https://github.com/holepunchto/hyperbee/blob/48e2735a857bde5561c4db0495f477d36ac143c0/index.js#L152)
- [`Hyperblobs`'s prefetcher `_update`](https://github.com/holepunchto/hyperblobs/blob/a7c2595584be8a2e7a98643a768013912cc6bca7/lib/prefetcher.js#L49)

### DIFF
--- a/index.js
+++ b/index.js
@@ -554,6 +554,7 @@ class Hypercore extends EventEmitter {
     this.core.replicator.findingPeers -= this._findingPeers
     this.core.replicator.clearRequests(this.activeRequests, error)
     this.core.replicator.updateActivity(this._active ? -1 : 0)
+    await this.core.replicator.localBitfield.flush()
 
     this._findingPeers = 0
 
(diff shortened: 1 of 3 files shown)


lib/core.js
```diff
@@ -669,11 +643,9 @@ module.exports = class Core {
 
     const { start, length, drop } = bitfield
 
-    this._setBitfieldRanges(start, start + length, true)
-    this.updateContiguousLength({ start, length, drop: false })
-
-    this.replicator.onupgrade()
     this.replicator.onhave(start, length, drop)
+    this.replicator.onupgrade()
```
mafintosh: Should be before onhave so it doesn’t signal out of bounds blocks

PR #743, opened 2025-10-27
Add `core.remoteContiguousLength` & `remote-contiguous-length` event

lejeunerenard:
Add the property `.remoteContiguousLength` for the maximum known peer contiguous length and the `remote-contiguous-length` event for when it updates.

Added documentation noting the following exceptions:
- The remote contiguous length is not updated when a peer is not fully contiguous.
- The event is not fired due to truncation.

### DIFF
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -1163,7 +1163,7 @@ class Peer {
       if (length > this._remoteContiguousLength) {
         this._remoteContiguousLength = length
         if (
-          this.remoteFork >= this.core.state.fork &&
+          this.remoteFork === this.core.state.fork &&
           length > this.core.header.hints.remoteContiguousLength
         ) {
           await this.core.updateRemoteContiguousLength(length)


lib/core.js
```diff
@@ -852,7 +864,8 @@ function parseHeader(info) {
     tree: info.head || getDefaultTree(),
     hints: {
       reorgs: [],
-      contiguousLength: info.hints ? info.hints.contiguousLength : 0
+      contiguousLength: info.hints ? info.hints.contiguousLength : 0,
+      remoteContiguousLength: info.hints ? info.hints.remoteContiguousLength : 0
```
HDegroote, line 870: Might be irrelevant in practice, but in theory `info.hints` can be set but without both of these fields present, since they're optional, so something like `remoteContiguousLength: info.hints?.remoteContiguousLength ? info.hints.remoteContiguousLength : 0` seems safer

lib/replicator.js
```diff
@@ -513,18 +513,18 @@ class Peer {
     else this._clearLocalRange(start, length)
 
     const i = Math.floor(start / DEFAULT_SEGMENT_SIZE)
-    const contig = this.core.header.hints.contiguousLength === this.core.state.length
+    const fullyContig = this.core.header.hints.contiguousLength === this.core.state.length
```
HDegroote, line 517: (detail) unneeded renaming of the variable?

lib/replicator.js
```diff
@@ -1156,6 +1156,12 @@ class Peer {
       this._remoteContiguousLength = start
     }
 
+    const storedRemoteContig = this.core.header.hints.remoteContiguousLength
+
+    if (drop === false && start + length > storedRemoteContig && start <= storedRemoteContig) {
```
HDegroote: I don't think this logic is correct: we should never update the `remoteContiguousLength`for a peer who has the core sparsely (so a range that doesn't start with 0).

Fixed by putting this logic inside the if just below.

(should be easy to check with a test)

index.js
```diff
@@ -561,6 +561,11 @@ class Hypercore extends EventEmitter {
     return this.state.byteLength - this.state.length * this.padding
   }
 
+  get remoteContiguousLength() {
+    if (this.opened === false) return 0
+    return Math.min(this.core.state.length, this.core.header.hints.remoteContiguousLength)
```
HDegroote, line 569: (question) Why is it needed to take this min?

lib/replicator.js
lejeunerenard: The renaming was because initially I thought it meant that the local core was contiguous even if partially so. So renamed only to clarify that it is fully contiguous in the sense that it matches the latest length.

But yeah, no functionality change.

index.js
lejeunerenard: To account for truncation, for example. Same logic is applied for the `contiguousLength` as well.

If the state length is less then any hint, then that's the extent that can be asserted.

lib/core.js
lejeunerenard: Good point. Especially the `remoteContiguousLength` could be undefined if used with an older `hypercore-storage` version.

I updated and also used nullish coalescing to make it easier to read.

lib/replicator.js
lejeunerenard: @HDegroote I thought we would allow it to sparsely update, but after reviewing the code again, we never send partial ranges from contiguous peers, right? So you're right we should change this since a partial range would imply the peer was actually not contiguous.

HDegroote: Yes, if you are contiguous you always send a range starting at 0. You only send a range that doesn't start at 0 if there's a gap in the blocks you have, for example range(0,100) and range(102,200) if you miss block 101. You wouldn't send range(0,100) and range(101,200)

HDegroote: (there's no benefit in sparsely updating it, since a range is always just 2 integers: start and end)

lejeunerenard: Done here (b4c21dad411b0d9c1ae8db812d366d458fb3959a) Nice consequence is that the code simplifies to only calling `core.updateRemoteContiguousLength()`.

lib/replicator.js
```diff
@@ -1157,7 +1157,10 @@ class Peer {
     }
 
     if (start === 0 && drop === false) {
-      if (length > this._remoteContiguousLength) this._remoteContiguousLength = length
+      if (length > this._remoteContiguousLength) {
+        this._remoteContiguousLength = length
+        this.core.updateRemoteContiguousLength(length).catch(noop)
```
chm-diederichs: I'm worried about not handling the error here cause we are flushing to storage

lib/session-state.js
```diff
@@ -473,13 +473,37 @@ module.exports = class SessionState {
     if (this.isDefault()) {
       await storeBitfieldRange(this.storage, storage, batch.ancestors, batch.treeLength, false)
       if (batch.ancestors < this.core.header.hints.contiguousLength) {
-        storage.setHints({ contiguousLength: batch.ancestors })
+        this.core.header.hints.remoteContiguousLength = Math.min(
+          batch.length,
```
chm-diederichs, line 477: Unsure if it makes a difference except on reorgs, but should we use `batch.ancestors` here to align with `contiguousLength`?

lib/replicator.js
lejeunerenard: Was leftover from the WIP. I agree that it is concerning and that any error should bubble up. Will remove momentarily.

lib/replicator.js
```diff
@@ -1157,7 +1157,10 @@ class Peer {
     }
 
     if (start === 0 && drop === false) {
-      if (length > this._remoteContiguousLength) this._remoteContiguousLength = length
+      if (length > this._remoteContiguousLength) {
+        this._remoteContiguousLength = length
+        this.core.updateRemoteContiguousLength(length)
```
chm-diederichs: I'm not sure we can let it go uncaught either.

I think either we need a dedicated error handler, but we don't have any standard place to emit the event.. 

What do you think about changing `updateRemoteContiguousLength` to be sync and have it set a `needsHintFlush` flag on the core, then `SessionState.flush` always checks that flag and when flushed [these lines](https://github.com/holepunchto/hypercore/pull/743/files#diff-826e5ed9069c92baa50ec46dde8074a6789b4c90837c5ede0e6b0438e3c34309R401-R403) are triggered by calling a `_onhintflush` hook instead?

chm-diederichs: it would mean that `remoteContiguousLength` may be stale on restart, not sure of the ramifications there

chm-diederichs: seems like we aren't bubbling up from the replicator anyways, so can just catch with an error handler that mimics [here](https://github.com/holepunchto/hypercore/blob/f11a61cd3bea490637825af05ec12b74ec5ef15c/lib/replicator.js#L1018-L1021)

lejeunerenard: @chm-diederichs so a safetycatch and is the `WRITE_FAILED` the only error code we want to pause from?

chm-diederichs: no I think we should pause on any error, my link was misleading, I meant to refer to just the safetyCatch and the pause

lib/session-state.js
lejeunerenard: @chm-diederichs I used `length` because I thought this would cover cases where a peer has a fork with further blocks after the truncation point. For the contiguous length, it makes sense to use the ancestor because the rest of the blocks need to be updated and verified etc. For the remote contiguous length, are we guaranteed a range request after a reorg that would allow the length to be updated to include the newer blocks?

lejeunerenard: Added this test to illustrate the point:
https://github.com/holepunchto/hypercore/pull/743/commits/e8ef422af2e1c091cd97ef21a7c25c6217ec34dd
The test will fail [here](https://github.com/holepunchto/hypercore/pull/743/commits/e8ef422af2e1c091cd97ef21a7c25c6217ec34dd#diff-83ae9e94eb595eb1f5698b58bc4f55ae66a6192fd1f69bf4d6a764ce8c03b165R2529) when using `batch.ancestors` but works with `batch.length`.

Also note that this test will fail regardless ([here](https://github.com/holepunchto/hypercore/pull/743/commits/e8ef422af2e1c091cd97ef21a7c25c6217ec34dd#diff-83ae9e94eb595eb1f5698b58bc4f55ae66a6192fd1f69bf4d6a764ce8c03b165R2527)) because it looks like the logic around using the local state length for the `.remoteContiguousLength` does seem to have a bug.

lejeunerenard: Fixed the writer's remoteContiguousLength being too high in https://github.com/holepunchto/hypercore/pull/743/commits/39af9e503e1c08a3530c12fcb8a11fbf0377db58 by checking the fork of the remote before updating the `.remoteContiguousLength`. The range request could come after the truncation and would update the length.

lib/replicator.js
lejeunerenard: Fixed in https://github.com/holepunchto/hypercore/pull/743/commits/52a3e890385ce516faf710704e69d447b9a262c4

lib/session-state.js
chm-diederichs: ok nice I follow now, thanks for clarifying

lib/core.js
```diff
@@ -391,6 +392,17 @@ module.exports = class Core {
     return this.state.flushed()
   }
 
+  async updateRemoteContiguousLength(length) {
```
chm-diederichs: just a nit, but could we move this down to live next to `updateContiguousLength`

lib/core.js
```diff
@@ -855,7 +867,8 @@ function parseHeader(info) {
     tree: info.head || getDefaultTree(),
     hints: {
       reorgs: [],
-      contiguousLength: info.hints ? info.hints.contiguousLength : 0
+      contiguousLength: info.hints?.contiguousLength ?? 0,
+      remoteContiguousLength: info.hints?.remoteContiguousLength ?? 0
```
mafintosh: why info.hints? is hints not always defined?

lib/replicator.js
```diff
@@ -1152,15 +1152,26 @@ class Peer {
...
 
     if (drop === true && start < this._remoteContiguousLength) {
       this._remoteContiguousLength = start
     }
 
     if (start === 0 && drop === false) {
-      if (length > this._remoteContiguousLength) this._remoteContiguousLength = length
+      if (length > this._remoteContiguousLength) {
+        this._remoteContiguousLength = length
+        if (this.remoteFork >= this.core.state.fork) {
```
mafintosh: would add this condition to the if (inverse) https://github.com/holepunchto/hypercore/pull/743/files#diff-826e5ed9069c92baa50ec46dde8074a6789b4c90837c5ede0e6b0438e3c34309R718

lib/replicator.js
```diff
@@ -1152,15 +1152,26 @@ class Peer {
...
+          try {
+            await this.core.updateRemoteContiguousLength(length)
```
mafintosh: dont try catch it, let it bubble up. if this fails, catastrophic things are happening (disk on fire)

lib/core.js
lejeunerenard: Yes in practice, `info.hints` is always defined. Why was `info.hints` checked originally for `contiguousLength` here? https://github.com/holepunchto/hypercore/commit/d52a6f4589d58e65646d3942851ac4b65666084a
Since the hints come from `hypercore-storage`, it might have updated since then?

lib/replicator.js
lejeunerenard: Fixed in fb95516d13272567e0b7bef2609e540e4f6a1311

lib/replicator.js
lejeunerenard: Done in 925dadfd559928ec8fdea550295cceb5d88911e7

lib/core.js
mafintosh: ah ok, i prefer the other style of checking personally

PR #745, opened 2025-11-28
eagerly flush hints

mafintosh:


lib/core.js
```diff
@@ -539,6 +543,20 @@ module.exports = class Core {
     return verifies[0] !== null
   }
 
+  async flushHints() {
```
HDegroote, line 546: ATM there are race conditions possible. It's trivially solved by debouncing this function and setting `this.hintsChanged = false` right after the early-return check (in the same tick).

mafintosh: fixed :)

HDegroote: Clarified out of band that this runs in a lock, so we are fully protected from race conditions by setting `hintsChanged = false` in the same tick

PR #747, opened 2025-12-10
if a range supports it, request future blocks in one go when requesti…

mafintosh:
…ng proof

### DIFF
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -1274,9 +1274,15 @@ class Peer {
 
     const index = this.remoteLength - 1
 
+    // only valid if its an OOB get
+    if (index < this.core.state.length) return null
+
+    // do the normal checks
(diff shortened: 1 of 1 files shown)


lib/replicator.js
```diff
@@ -1843,10 +1873,8 @@ module.exports = class Replicator {
     // Also auto compresses the range based on local bitfield
     clampRange(this.core, r)
 
-    this._ranges.push(r)
```
lejeunerenard, line 1846: Was this accidental? This is the only place where we add a range to `this._ranges`, no?

mafintosh: i moved it to the constructor of the range request, much cleaner imo

lejeunerenard: 🤦‍♂️

lejeunerenard (approved):

PR #749, opened 2025-12-10
Allow push

mafintosh:


lib/replicator.js
```diff
@@ -937,6 +976,12 @@ class Peer {
       this.replicator._onupload(proof.block.index, proof.block.value.byteLength, this)
     }
 
+    if ((pushing || this.pushing) && proof.upgrade) {
+      this.pushing = true
+      const remoteLength = proof.upgrade.start + proof.upgrade.length
+      if (remoteLength > this.remoteLength) this.remoteLength = remoteLength
+    }
```
lejeunerenard, line 1004: This will always be allowed to run if theres an upgrade after one push since `this.pushing = true`. Do we want to update `remoteLength` for non-push requests too? Is it safe to assume the upgrade will be applied?

lib/replicator.js
```diff
@@ -1062,7 +1124,11 @@ class Peer {
 
   onnodata({ request, reason }) {
     const req = request > 0 ? this.replicator._inflight.get(request) : null
-    if (req === null || req.peer !== this) return
+
+    if (req === null || req.peer !== this) {
+      this.updateAll()
```
```suggestion
      this.replicator.updateAll()
```

lib/replicator.js
mafintosh: as long as its in push mode we should. this is a little bit ... iffy tbh, but its to help predict the right proof to produce (similar to your invalid op comment) - there is a better way with a sep property, let me add a comment. we should also do it pased on remoteAllowPush now

mafintosh: ah we already do implicly cause we guard with remoteAllowPush in general, but let me comment

lejeunerenard (approved):
➡️ 📦

PR #758, opened 2026-01-13
Make want batches smaller

mafintosh:
wip

### DIFF
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -1685,7 +1685,7 @@ class Peer {
     return false
   }
 
-  _populateRangeBatches(r, debug) {
+  _populateRangeBatches(r) {
     for (let i = r.batches.length - 1; i >= 0; i--) {
       const b = r.batches[i]
       const minStart = b.batch * WANT_BATCH


lib/replicator.js
```diff
@@ -118,7 +128,37 @@ class Attachable {
...
+        const removed = w.wants.removeAnyRange(w)
+        if (!removed) continue
+        if (update === null) update = new Set()
+        update.add(w.wants.peer)
+        // TODO: should also use a free list for simplicity
+        w.wants.peer.wireUnwant.send(removed)
+        continue
+      }
+
+      if (w.wants.removeBatch(w.index, this)) {
```
```suggestion
      if (w.wants.removeBatch(w.start, this)) {
```

Also implies we aren't testing that the wants are removed for cancelled requests.

lib/replicator.js
```diff
@@ -118,7 +128,37 @@ class Attachable {
...
+    if (this.wants === null) return
+
+    const wants = this.wants
+    this.wants = null
+
+    let update = null
+
+    for (const w of wants) {
+      if (w.any) {
+        const removed = w.wants.removeAnyRange(w)
```
```suggestion
        const removed = w.wants.removeAnyRange(w.start, w.length, this)
```

lib/replicator.js
```diff
@@ -1547,68 +1635,114 @@ class Peer {
...
-      if (min > -1) this._maybeWant(min, max - min)
-      return false
+  _populateRangeBatches(r) {
+    for (let i = r.batches.length - 1; i >= 0; i--) {
+      const b = r.batches[i]
+      const sparse = this.core.bitfield.hasUnset(b * WANT_BATCH, WANT_BATCH)
+      if (sparse) continue
+      const head = r.batches.pop()
+      if (head !== b) r.batches[i] = head
+      this.wants.remove(b * WANT_BATCH, r)
```
```suggestion
      this.wants.removeBatch(b, r)
```

lib/replicator.js
```diff
@@ -1180,15 +1233,13 @@ class Peer {
...
-    this.replicator._onwant(this, start, length)
+  onwant(range) {
+    if (!this.remoteWants.add(range)) return
+    this.replicator._onwant(this, range.start, range.length, range.any)
   }
 
-  onunwant() {
-    // TODO
+  onunwant(range) {
+    this.remoteWants.delete(range)
```
```suggestion
    this.remoteWants.remove(range)
```

lib/wants.js
```diff
@@ -0,0 +1,298 @@
...
+
+    if (smallest === -1) return this.all
+
+    const r = start & (smallest - 1)
+    const end = start + length
+
+    let max = 3
+
+    for (let i = start - r; i < end; i += smallest) {
+      if (max-- === 0) return true // just to save work
```
lejeunerenard, line 190:
```suggestion
    let max = MAX_REMOTE_BATCHES

    for (let i = start - r; i < end; i += smallest) {
      if (--max === 0) return true // just to save work
```
Might make this a bit clearer.

lib/wants.js
```diff
@@ -0,0 +1,298 @@
...
+
+    if (this.batches.length >= MAX_REMOTE_BATCHES) {
+      this.all = true
+      return true
+    }
+
+    this.batches.push({
+      length: range.length,
+      ranges: new Set([range.start / range.length])
+    })
```
lejeunerenard, line 244:
```suggestion
    })
    this.size++
```

I assume that `.size` is the total number of ranges. Especially since if you add a single range and remove it, the `size` goes to `-1` without the above.

lib/wants.js
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    if (!validateBatchRange(range)) return false
+
+    for (let i = 0; i < this.batches.length; i++) {
+      const b = this.batches[i]
+      if (b.length !== range.length) continue
+
+      const size = b.ranges.size
+      b.ranges.delete(range.start)
```
lejeunerenard:
```suggestion
      b.ranges.delete(range.start / range.length)
```
`unwant` messages are the block index not batch index.

lib/replicator.js
```diff
@@ -2565,14 +2679,22 @@ module.exports = class Replicator {
...
+        peer.wireRange.send({
+          drop: false,
+          start: bit,
+          length: 1
+        })
+      }
+      return
+    }
 
     for (const msg of this.core.bitfield.want(start, length)) {
```
lejeunerenard, line 2734: Why is responding to the `want` with bitfield messages no longer corked?

mafintosh: ah cause its its 1 in practice now

lib/wants.js
mafintosh: it shouldnt be --max, it should be max--  otherwise its only max - 1 runs

lejeunerenard: Yeah `--max` is more confusing and less common. The main thing I wanted to clarify is why `3` attempts? From what I understand it's because of the max number of remote batches. It's one less because if we get to the max number of batches [like we do here](https://github.com/holepunchto/hypercore/pull/758/files#diff-9e07ab4141163824d90ea4b3accae0057085f8b594c3f7cb364d7662f335f899R228) then the results is stored in the `.all` property. So maybe this makes more sense? 
```suggestion
    let max = MAX_REMOTE_BATCHES - 1

    for (let i = start - r; i < end; i += smallest) {
      if (max-- === 0) return true // just to save work
```

lib/replicator.js
```diff
@@ -1547,68 +1670,119 @@ class Peer {
...
-      }
+      offset = (b + 1) * WANT_BATCH
+    }
 
-      if (min > -1) this._maybeWant(min, max - min)
-      return false
+    return false
+  }
+
+  _populateRangeBatches(r, debug) {
```
```suggestion
  _populateRangeBatches(r) {
```

lejeunerenard (approved):

PR #767, opened 2026-02-04
Merkle tree recovery

lejeunerenard:
First step is to allow missing roots when `ready`ing a hypercore. The roots were used to determine the length based on their span. If the roots are not available the header's tree `length` value is used.

The general technique to recover a merkle tree node, is to pass a fully remote proof from a peer with the node to the affected peer.

1. A fully remote proof can be generated targeting the node (via it's merkle index which is not equivalent but related to it's block index) by calling:
   ```js
   const proof = await core.generateRemoteProofForTreeNode(treeNodeIndex)
   ```
2. The proof can then be verified and applied on the peer missing the node like so:
   ```js
   await core.recoverFromRemoteProof(proof)
   ```

Since proofs include the nodes as part of the proof, the peer without the node can directly write the nodes after verifying them. A new `upgrade` argument was added to `fully-remote-proof.js`'s `proof()` to support overriding the default `upgrade` for the sender's `length`. This allows the proof generated for recovery to target subtree roots.

Tests were added to demonstrate this flow for both root nodes and subtree root nodes. A new assertion was added for the `fully-remote-proof.js` tests as well for the added `upgrade` argument.

### DIFF
--- a/test/merkle-tree-recovery.js
+++ b/test/merkle-tree-recovery.js
@@ -60,8 +60,9 @@ test('recover - bad merkle root - fix via fully remote proof', async (t) => {
   // Delete tree nodes
   const tx = core.core.storage.write()
   const [rootIndex] = flat.fullRoots(2 * num)
+  const targetBlockIndex = flat.rightSpan(rootIndex) / 2 + 1
 
-  const initialHash = await core.treeHash(rootIndex) // store for later check
+  const initialHash = await core.treeHash(targetBlockIndex) // store for later check
 
   // Get proof from good core, before deleting
   const p = await core.generateRemoteProofForTreeNode(rootIndex)
@@ -84,8 +85,7 @@ test('recover - bad merkle root - fix via fully remote proof', async (t) => {
 
   t.is(core2.length, num, 'still has length')
 
-  const hash = await core2.treeHash(rootIndex)
-  t.alike(hash, initialHash, 'still can construct the hash')
+  await t.exception(core2.treeHash(targetBlockIndex), 'INVALID_OPERATION', 'cant create tree hash')
 
   // Verify remote proof & patch with it's proof
   t.ok(await core2.recoverFromRemoteProof(p), 'recovery verified correctly')
@@ -186,8 +186,9 @@ test('recover - bad merkle root - fix via range request to include roots automat
   // Delete tree nodes
   const tx = core.core.storage.write()
   const [rootIndex] = flat.fullRoots(2 * num)
+  const targetBlockIndex = flat.rightSpan(rootIndex) / 2 + 1
 
-  const initialHash = await core.treeHash(rootIndex) // store for later check
+  const initialHash = await core.treeHash(targetBlockIndex) // store for later check
 
   // Get proof from good core, before deleting
   t.ok(await MerkleTree.get(core.core, rootIndex))
@@ -223,8 +224,7 @@ test('recover - bad merkle root - fix via range request to include roots automat
 
   t.is(core2.length, num, 'still has length')
 
-  const hash = await core2.treeHash(rootIndex)
-  t.alike(hash, initialHash, 'still can construct the hash')
+  await t.exception(core2.treeHash(targetBlockIndex), 'INVALID_OPERATION', 'cant create tree hash')
 
   const [repairingProof] = await repairing
   t.is(repairingProof.upgrade.length, 30, 'repairing emitted')


lib/core.js
```diff
@@ -606,37 +606,7 @@ module.exports = class Core {
...
-        const treeProof = await treeProofPromise
-
-        const verifyBatch = MerkleTree.verifyFullyRemote(this.state, await treeProof.settle())
-        this._verifyBatchUpgrade(verifyBatch, this.header.manifest)
-      } catch (err) {
-        return true
-      }
-
-      return false
+      return !this._repairTreeNodes(proof)
```
mafintosh: await

lib/core.js
```diff
@@ -265,13 +265,18 @@ module.exports = class Core {
 
     const bitfield = await Bitfield.open(storage, header.tree.length)
 
+    let roots = []
+    try {
+      roots = header.tree.length
+        ? await MerkleTree.getRootsFromStorage(storage, header.tree.length)
+        : []
+    } catch {}
```
mafintosh: lets throw if not invalid op

lib/core.js
```diff
@@ -573,7 +578,10 @@ module.exports = class Core {
       return true
     }
 
-    const roots = await MerkleTree.getRootsFromStorage(this.storage, proof.upgrade.length)
+    let roots = []
+    try {
+      roots = await MerkleTree.getRootsFromStorage(this.storage, proof.upgrade.length)
+    } catch {}
```
mafintosh: same as above, mb move to a function that returns roots? cleaner imo and then can be reused

lib/fully-remote-proof.js
```diff
@@ -65,7 +65,7 @@ async function verify(storage, buffer, { referrer = null } = {}) {
 
   rx.tryFlush()
 
-  const roots = await Promise.all(rootPromises)
+  const roots = (await Promise.all(rootPromises)).filter(Boolean)
```
mafintosh: this needed? scary

lejeunerenard: Yeah if the core is missing a node, then that promise will return `null`. If a core that's repairing tries to verify a proof with nodes it doesn't contain, it would throw trying to verify a `null` node. Given what roots it has are used to verify the proof and its signature is tested as well, I thought that was sufficient. I.E. we test it as much as we can check with the knowledge we have.

lib/core.js
lejeunerenard: New function here:
https://github.com/holepunchto/hypercore/pull/767/changes#diff-826e5ed9069c92baa50ec46dde8074a6789b4c90837c5ede0e6b0438e3c34309R955-R963

lib/core.js
```diff
@@ -899,6 +963,16 @@ function parseHeader(info) {
   }
 }
 
+async function _maybeGetRootsFromStorage(storage, length) {
```
mafintosh: nit but functions in this scope should never start with _

lib/fully-remote-proof.js
mafintosh: better to not do that and either reject all local knowledge ie, pretend length = 0 roots = [] or throw - this might work atm but leads to a state machine where an attacker might try to do quirky stuff to get a bad proof passed

lib/fully-remote-proof.js
```diff
@@ -65,7 +65,9 @@ async function verify(storage, buffer, { referrer = null } = {}) {
 
   rx.tryFlush()
 
-  const roots = await Promise.all(rootPromises)
+  let roots = await Promise.all(rootPromises)
+  if (roots.some((r) => r === null)) roots = []
```
mafintosh: cache the function ie, isNull(r) { return r === null } at bottom of file

lib/merkle-tree.js
```diff
@@ -231,6 +231,11 @@ class MerkleTreeBatch {
 
   commit(tx) {
     if (tx === undefined) throw INVALID_OPERATION('No database batch was passed')
+
+    if (this.session.core._repairMode) {
+      throw Error('Cannot commit while repair mode is on')
```
mafintosh: be nice to use a typed error from hypercore-errors

lejeunerenard: Used `ASSERTION` and changed the `assert()` in `lib/session-state.js` to use the same for the same error. These will ensure its thrown as uncaughts by `safety-catch`.

PR #770, opened 2026-02-16
only update pushedLength when pushing

chm-diederichs:


test/replicate.js
```diff
@@ -2835,6 +2835,36 @@ test('no requests in pushOnly mode', async function (t) {
   await t.execution(b.get(1, { timeout: 500 }))
   await t.execution(b.get(2, { timeout: 500 }))
 })
+
+test('push and pull concurrently', async function (t) {
+  const a = await create(t)
+  const b = await create(t, a.key, { allowPush: true, pushOnly: true })
+
+  b.replicator.setPushOnly(true)
```
lejeunerenard: Redundant with the `opts` passed to the constructor above.

test/replicate.js
```diff
@@ -2835,6 +2835,36 @@ test('no requests in pushOnly mode', async function (t) {
...
+  for (let i = 0; i < 20; i++) {
+    await a.append(i.toString())
+  }
+
+  const appends = []
+  for (let i = 20; i < 30; i++) {
+    appends.push(a.append(i.toString()).then(() => a.replicator.push(i)))
+  }
+
+  await 1
```
lejeunerenard:
```suggestion
```

test/replicate.js
```diff
@@ -2835,6 +2835,36 @@ test('no requests in pushOnly mode', async function (t) {
...
+  await b.get(10, { force: true, timeout: 200 })
+
+  await Promise.all(appends)
+  await new Promise((resolve) => setTimeout(resolve, 500))
```
lejeunerenard: Do we need to use timeouts? Maybe to ensure blocks are pushed quickly? Otherwise it use the `append` event on `b` to precisely check `b`'s `length` w/o timeouts.

lib/replicator.js
```diff
@@ -1045,25 +1050,31 @@ class Peer {
...
+      } else {
+        req.upgrade === null
+        this.replicator._txLock.unlock()
+        locked = false
+      }
 
-      // If cutting off the fulfillment, ensure promise is caught.
-      // While repairing merkle tree nodes can be `nil` so can throw.
-      req.fulfill().catch(safetyCatch)
-      return
```
lejeunerenard, line 1064: The repair mode guard looks like it was accidentally removed. Likely because the push mode guard and this are almost the same.

lib/replicator.js
```diff
@@ -1045,25 +1050,31 @@ class Peer {
 
     batch.tryFlush()
 
-    await this._fulfillRequest(req, false)
-  }
+    if (!req.upgrade) {
+      return this._fulfillRequest(req, false)
```
lejeunerenard: This currently always runs as `ProofRequest` doesn't have an `upgrade` property. Maybe `req` should be `req.proof`? But then the modifications to the `upgrade` would be moot given the upgrade nodes have been queued already.

lib/replicator.js
```diff
@@ -1045,25 +1050,31 @@ class Peer {
...
+    try {
+      locked = true
+      const end = req.upgrade.start + req.upgrade.length
 
-      // sync from now on, so safe to delete from the map
-      this.remoteRequests.delete(req.msg.id)
+      if (this.pushedLength < end) {
+        req.upgrade.start = this.pushedLength
+      } else {
+        req.upgrade === null
```
mafintosh: you mean = ?

lib/replicator.js
chm-diederichs: ahh my bad, i saw this at one point but thought it was an upstream diff from rebasing. will update now

lib/replicator.js
mafintosh: makes sense, we should move the upgrade adjustment and have it mutate the message i guess

chm-diederichs: or just `msg.upgrade`?

lib/replicator.js
```diff
@@ -1035,36 +1035,34 @@ class Peer {
...
   async _handleRequest(msg) {
-    const batch = this.core.storage.read()
+    const locked = !!msg.upgrade && this.remoteAllowPush
 
-    // TODO: could still be answerable if (index, fork) is an ancestor of the current fork
-    const req =
-      msg.fork === this.core.state.fork
-        ? await this._getProof(batch, msg, false)
-        : new ProofRequest(msg, null, null, null)
+    if (this.remoteAllowPush && locked) { // ensure contig push proofs
```
chm-diederichs: just `locked` would do, it implies this.remoteAllowPush

mafintosh: good point, fixed

lib/replicator.js
lejeunerenard: Fixed.

PR #772, opened 2026-02-19
Bail push if index is greater than length

chm-diederichs:


test/push.js
```diff
@@ -0,0 +1,87 @@
...
+  await b.get(10, { force: true })
+
+  await Promise.all(appends)
+  await bHasLength
+
+  t.pass('b synced length')
+  t.ok(await b.has(29))
+})
+
+test.solo('push before append', async function (t) {
```
mafintosh: no solo

chm-diederichs: i left it there so we address the test before merging, it's very timing dependent

test/push.js
lejeunerenard, line 1: Like the idea of moving to it's own test file! 🗂️  Don't forget to add to `test/all.js`.

PR #798, opened 2026-04-21
Use Brittle v4

jkcdarunday:


jkcdarunday:
Replaced by https://github.com/holepunchto/hypercore/pull/806

PR #809, opened 2026-05-14
Add close-path replication reschedule coverage

marcus-pousette-hp:
Adds coverage for peer-close scheduling paths:

- inflight block requests resume on a remaining peer after the requested peer closes
- pending range downloads resume on a remaining peer after an idle peer closes

This PR is test-only. The idle-close updateAll() optimization was moved to a follow-up branch.

### DIFF
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -2465,8 +2465,16 @@ module.exports = class Replicator {
 
     this._onpeerupdate(false, peer)
 
-    // If the peer had no requests to clear, there is nothing to reschedule.
-    if (needsUpdate && this.peers.length > 0) this.updateAll()
+    // A close can also wake pending work that was not assigned to the closing peer.
+    const hasPendingUpdate =
+      needsUpdate ||
+      this._queued.length > 0 ||
+      this._seeks.length > 0 ||
+      this._ranges.length > 0 ||
+      this._reorgs.length > 0 ||
+      this._maybeUpdate()
+
+    if (hasPendingUpdate && this.peers.length > 0) this.updateAll()
     else this._maybeResolveAvailable()
   }
 
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -754,7 +754,7 @@ class Peer {
 
     if (this.remoteOpened === false) {
       this.replicator._ifAvailable--
-      this.replicator.updateAll()
+      this.replicator._maybeResolveAvailable()
       return
     }
 
@@ -2449,16 +2449,25 @@ module.exports = class Replicator {
     this.peers.splice(this.peers.indexOf(peer), 1)
     peer.wants.destroy()
 
-    if (this._manifestPeer === peer) this._manifestPeer = null
+    let needsUpdate = false
+
+    if (this._manifestPeer === peer) {
+      this._manifestPeer = null
+      needsUpdate = true
+    }
 
     for (const req of this._inflight) {
       if (req.peer !== peer) continue
       this._inflight.remove(req.id, true)
       this._clearRequest(peer, req)
+      needsUpdate = true
     }
 
     this._onpeerupdate(false, peer)
-    this.updateAll()
+
+    // If the peer had no requests to clear, there is nothing to reschedule.
+    if (needsUpdate && this.peers.length > 0) this.updateAll()
+    else this._maybeResolveAvailable()
   }
 
   _queueBlock(b) {
@@ -2666,6 +2675,11 @@ module.exports = class Replicator {
         i--
       }
     }
+  }
+
+  _maybeResolveAvailable() {
+    this._checkUpgradeIfAvailable()
+    this._maybeResolveIfAvailableRanges()
   }
 
   _clearRequest(peer, req) {
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -754,7 +754,7 @@ class Peer {
 
     if (this.remoteOpened === false) {
       this.replicator._ifAvailable--
-      this.replicator._maybeResolveAvailable()
+      this.replicator.updateAll()
       return
     }
 
@@ -2449,33 +2449,16 @@ module.exports = class Replicator {
     this.peers.splice(this.peers.indexOf(peer), 1)
     peer.wants.destroy()
 
-    let needsUpdate = false
-
-    if (this._manifestPeer === peer) {
-      this._manifestPeer = null
-      needsUpdate = true
-    }
+    if (this._manifestPeer === peer) this._manifestPeer = null
 
     for (const req of this._inflight) {
       if (req.peer !== peer) continue
       this._inflight.remove(req.id, true)
       this._clearRequest(peer, req)
-      needsUpdate = true
     }
 
     this._onpeerupdate(false, peer)
-
-    // A close can also wake pending work that was not assigned to the closing peer.
-    const hasPendingUpdate =
-      needsUpdate ||
-      this._queued.length > 0 ||
-      this._seeks.length > 0 ||
-      this._ranges.length > 0 ||
-      this._reorgs.length > 0 ||
-      this._maybeUpdate()
-
-    if (hasPendingUpdate && this.peers.length > 0) this.updateAll()
-    else this._maybeResolveAvailable()
+    this.updateAll()
   }
 
   _queueBlock(b) {
@@ -2683,11 +2666,6 @@ module.exports = class Replicator {
         i--
       }
     }
-  }
-
-  _maybeResolveAvailable() {
-    this._checkUpgradeIfAvailable()
-    this._maybeResolveIfAvailableRanges()
   }
 
   _clearRequest(peer, req) {
--- a/test/replicate.js
+++ b/test/replicate.js
@@ -899,13 +899,10 @@ test('closing idle peer schedules pending range on remaining peer', async functi
   const clone = await create(t, writer.key)
 
   const writerPeerWait = new Promise((resolve) => clone.once('peer-add', resolve))
-  const writerStreams = replicate(writer, clone, t, { teardown: false })
+  const writerStreams = replicate(writer, clone, t)
   const writerPeer = await writerPeerWait
 
-  const sparseStreams = replicate(sparse, clone, t, { teardown: false })
-
-  t.teardown(() => destroyStreams(writerStreams))
-  t.teardown(() => destroyStreams(sparseStreams))
+  const sparseStreams = replicate(sparse, clone, t)
 
   await clone.update({ wait: true })
   await eventFlush()
--- a/test/replicate.js
+++ b/test/replicate.js
@@ -793,6 +793,52 @@ test('multiplexing multiple times over the same stream', async function (t) {
   s3.destroy()
 })
 
+test('closing peer with inflight block reschedules on remaining peer', async function (t) {
+  const writer = await create(t)
+  await writer.append(['block-0', 'block-1', 'block-2'])
+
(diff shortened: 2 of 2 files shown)


test/replicate.js
```diff
@@ -793,6 +793,157 @@ test('multiplexing multiple times over the same stream', async function (t) {
...
+test('closing idle multiplexed stream does not update all peers', async function (t) {
+  const n1 = new NoiseSecretStream(true)
+  const n2 = new NoiseSecretStream(false)
+
+  n1.rawStream.pipe(n2.rawStream).pipe(n1.rawStream)
+
+  const cores = []
+  const clones = []
+
+  for (let i = 0; i < 4; i++) {
```
lejeunerenard: This should test multiple peers for one core, not multiple core pairs for two peers. With pairs of cores between two peers, the number of peers being checked on channel close are `0` because there's only one peer as you assert below.

lib/replicator.js
```diff
@@ -2449,16 +2449,33 @@ module.exports = class Replicator {
...
-    this.updateAll()
+
+    // A close can also wake pending work that was not assigned to the closing peer.
+    const hasPendingUpdate =
+      needsUpdate ||
+      this._queued.length > 0 ||
+      this._seeks.length > 0 ||
+      this._ranges.length > 0 ||
+      this._reorgs.length > 0 ||
+      this._maybeUpdate()
```
lejeunerenard: I'm not sure there's a difference any more between this set of criteria and just running `.updateAll()`. Is there a scenario where significant work is skipped when this check fails?

test/replicate.js
```diff
@@ -2894,6 +3045,73 @@ async function createAndDownload(t, core) {
...
 
+function findPeerForStream(core, remoteStream) {
+  for (const peer of core.core.replicator.peers) {
+    if (b4a.equals(peer.stream.remotePublicKey, remoteStream.publicKey)) return peer
+  }
+
+  throw new Error('peer not found')
+}
+
+async function waitForPeerForStream(core, remoteStream) {
```
lejeunerenard: Awaiting peers can be done via `once(core, 'peer-add')`. Usually this can be done in the correct order to get the peer you want in tests cases. Eg.

```js
const peerBWait = once(a, 'peer-add')
replicate(a, b, t)
const [peerBforA] = await peerBWait
```

Same concept can be used for `findPeerForStream` likely too.

test/replicate.js
```diff
@@ -793,6 +793,157 @@ test('multiplexing multiple times over the same stream', async function (t) {
...
+  const writer = await create(t)
+  const batch = []
+  for (let i = 0; i < 32; i++) batch.push('block-' + i)
+  await writer.append(batch)
+
+  const sparse = await create(t, writer.key)
+  const clone = await create(t, writer.key)
+
+  const writerStreams = replicate(writer, clone, t, { teardown: false })
+  const sparseStreams = replicate(sparse, clone, t, { teardown: false })
```
lejeunerenard: Why disable `teardown`?

test/replicate.js
```diff
@@ -2894,6 +3045,73 @@ async function createAndDownload(t, core) {
...
+      new Promise((resolve, reject) => {
+        timeout = setTimeout(() => reject(new Error(message)), ms)
+      })
+    ])
+  } finally {
+    clearTimeout(timeout)
+  }
+}
+
+function destroyStreams(streams) {
```
lejeunerenard: Same thing as `unreplicate()`.

test/replicate.js
```diff
@@ -793,6 +793,157 @@ test('multiplexing multiple times over the same stream', async function (t) {
...
+
+  const download = clone.download({ start: 0, end: writer.length })
+
+  await new Promise((resolve) => setTimeout(resolve, 100))
+  t.is(clone.core.replicator._inflight.idle, true, 'range has no inflight requests yet')
+
+  allowRange = true
+  await destroyStreams(sparseStreams)
+
+  await withTimeout(download.done(), 500, 'download did not resume after peer close')
```
lejeunerenard: Why does this need a timeout guard? Can't we jus await the promise.

test/replicate.js
```diff
@@ -793,6 +793,157 @@ test('multiplexing multiple times over the same stream', async function (t) {
...
+  let allowRange = false
+  const requestRange = writerPeer._requestRange
+  writerPeer._requestRange = function () {
+    if (!allowRange) return false
+    return requestRange.apply(this, arguments)
+  }
+
+  t.teardown(() => {
+    writerPeer._requestRange = requestRange
+  })
```
lejeunerenard: In my explorations, I pause the peer instead. `writerPeer.paused` set to `true` to stop requests and `false` to continue.

May be contrived as `paused` is changed with context, but is preferable to monkey patching which should be avoided as much as possible.

test/replicate.js
```diff
@@ -793,6 +793,157 @@ test('multiplexing multiple times over the same stream', async function (t) {
...
+
+  const download = clone.download({ start: 0, end: writer.length })
+
+  await new Promise((resolve) => setTimeout(resolve, 100))
+  t.is(clone.core.replicator._inflight.idle, true, 'range has no inflight requests yet')
+
+  allowRange = true
+  await destroyStreams(sparseStreams)
```
lejeunerenard: It would be good to assert that the sparse peer was closed via a `peer-remove` event. Not strictly necessary, but shows we are closing the stream to trigger the `onclose` event for the channel as the peer event is triggered from there.

test/replicate.js
```diff
@@ -793,6 +793,157 @@ test('multiplexing multiple times over the same stream', async function (t) {
...
+  const writer = await create(t)
+  await writer.append(['block-0', 'block-1', 'block-2'])
+
+  const mirror = await create(t, writer.key)
+  replicate(writer, mirror, t)
+  await mirror.download({ start: 0, end: writer.length }).done()
+
+  const clone = await create(t, writer.key)
+
+  const [slowWriterStream, slowCloneStream] = makeStreamPair(t, { latency: [200, 200] })
```
lejeunerenard: Does not need to be a slow stream if we await `upload` event on the peer we're getting the block from and close it then. Then it uses logical time.

test/replicate.js
```diff
@@ -793,6 +793,157 @@ test('multiplexing multiple times over the same stream', async function (t) {
...
+  let allowMirror = false
+  const requestBlock = mirrorPeer._requestBlock
+  mirrorPeer._requestBlock = function () {
+    if (!allowMirror) return false
+    return requestBlock.apply(this, arguments)
+  }
+
+  t.teardown(() => {
+    mirrorPeer._requestBlock = requestBlock
+  })
```
lejeunerenard: Same thing as below. We should prefer changing the peer's `.paused` over monkey patching.

test/replicate.js
```diff
@@ -793,6 +793,157 @@ test('multiplexing multiple times over the same stream', async function (t) {
...
+
+  const [slowWriterStream, slowCloneStream] = makeStreamPair(t, { latency: [200, 200] })
+  writer.replicate(slowWriterStream, { keepAlive: true })
+  clone.replicate(slowCloneStream, { keepAlive: true })
+
+  const mirrorStreams = replicate(mirror, clone, t, { keepAlive: true, teardown: false })
+  t.teardown(() => destroyStreams(mirrorStreams))
+
+  await clone.update({ wait: true })
+  await eventFlush()
```
lejeunerenard: Can be `await once(clone, 'append')` to await a length / metadata.

lejeunerenard (changes requested):
Good reproductions of the two test scenarios I mentioned out of band. With changes they will be less brittle & clearer.

I think though the fixes for when we call `.updateAll()` aren't netting us less work now that we have tests to show how it reschedules work potentially (ie the ranges test).

test/replicate.js
marcus-pousette-hp: Fixed https://github.com/holepunchto/hypercore/pull/809/commits/b8083d17009e6078e8ff2d3479ff207a717e36f5

lib/replicator.js
marcus-pousette-hp: Yeah. The story has changed a bit, given the new insights from the tests. 

Before we had something like 
```
close path was expensive, so avoid updateAll() on close unless needed.
```

now

```
avoid updateAll() only when the whole replicator is truly idle.
```

so it is much weaker/narrover improvement in general

marcus-pousette-hp: I rechecked the CPU profile. The close path is hot, but mostly because close triggers real pending range scheduling, not because it is doing empty idle scans.

The dominant stack is:

_removePeer -> updateAll -> _updatePeerNonPrimary -> _requestRange

That path is ~28% of the sampled CPU. So the safer version of this PR, which still calls updateAll() when ranges are pending, would not address the main profile hotspot. It may still help true idle close churn, but that is a much narrower optimization than the origanal purpose of the PR

Instead focus should be put more more strongly at the _requestRange / remote-bitfield / quickbit sparse-index path.

test/replicate.js
marcus-pousette-hp: Fixed https://github.com/holepunchto/hypercore/pull/809/commits/6a5faed794d47061ed63db73901d8f40735d8aa3

test/replicate.js
marcus-pousette-hp: That was leftover from an earlier temp-test cleanup issue with manually closing already-destroyed streams. The normal replicate() teardown handles this case correctly. I fixed this  https://github.com/holepunchto/hypercore/pull/809/commits/e61c1a009e5dab7d162bbe0d9facacdc2e409d7e

test/replicate.js
marcus-pousette-hp: Fixed https://github.com/holepunchto/hypercore/pull/809/commits/d2d0e2d289eaa36c87ea9d7077bc42dfeef366f6

test/replicate.js
marcus-pousette-hp: Good point. Much cleaner. Fixed https://github.com/holepunchto/hypercore/pull/809/commits/68cbe0c2b46bd0e5ef42bed029c5b6373247d73e

test/replicate.js
marcus-pousette-hp: Fixed https://github.com/holepunchto/hypercore/pull/809/commits/bd62b177a5d795e3900e797e2ba892c3a29011d7

test/replicate.js
marcus-pousette-hp: Yeap, that is way better. https://github.com/holepunchto/hypercore/pull/809/commits/b17110b5bb42f7d9f93a85b1a2d4c204d281233f

test/replicate.js
marcus-pousette-hp: Fixed https://github.com/holepunchto/hypercore/pull/809/commits/ad0a00f1a6f763b2269c8407df1704599116e6d1

test/replicate.js
marcus-pousette-hp: Bad cleanup by me, originally withTimeout(...) was useful while while I was exploring because a bad reschedule path could hang forever and the custom message made the failure obvious

marcus-pousette-hp: Fixed https://github.com/holepunchto/hypercore/pull/809/commits/af1c4d247304ae0ebc6454bed94f36c4d94d4883

lib/replicator.js
marcus-pousette-hp: I now split this PR into to. I repurposed this PR to be test only and https://github.com/holepunchto/hypercore/pull/812 for the potential minor gains

PR #828, opened 2026-06-01
feat: Include timestamp with group updates

Drache93:
Includes timestamps in the group events, also moves the event to post
flush to avoid races

### DIFF
--- a/lib/session-state.js
+++ b/lib/session-state.js
@@ -224,11 +224,12 @@ class SessionState {
 
     tx.deleteGroupUpdate(pointer, this.core.header.tree.timestamp)
     tx.putGroupUpdate(pointer, tree.timestamp, this.core.header.key)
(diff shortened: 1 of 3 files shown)


lib/session-state.js
```diff
@@ -224,11 +224,12 @@ class SessionState {
 
     tx.deleteGroupUpdate(pointer, this.core.header.tree.timestamp)
     tx.putGroupUpdate(pointer, tree.timestamp, this.core.header.key)
-    this.core.ongroupupdate(this.core.header.group.key, {
+    tx.pendingGroupUpdate = {
```
chm-diederichs: ahh let's not mutate the tx like this, 2 sec will edit

chm-diederichs (dismissed):

lejeunerenard (approved):

PR #832, opened 2026-06-22
Fix idle range completion scan

marcus-pousette-hp:
co-authored with AI

### DIFF
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -2602,13 +2602,14 @@ module.exports = class Replicator {
 
   _updateRanges(index, limit) {
     if (this._ranges.length === 0) {
-      return { checked: 0, resolved: 0, index: 0 }
+      return { checked: 0, resolved: 0, index: 0, updateAll: false }
     }
 
     index = Math.min(index, this._ranges.length - 1)
 
     let checked = 0
     let resolved = 0
+    let updateAll = false
     let remaining = Math.min(limit, this._ranges.length)
 
     while (remaining-- > 0 && this._ranges.length > 0) {
@@ -2622,6 +2623,8 @@ module.exports = class Replicator {
       if (r.end !== -1 && r.start >= r.end) {
         this._resolveRangeRequest(r)
         resolved++
+        // Crossing into the capped window exposes another range to peer scheduling.
+        if (this._ranges.length === MAX_RANGES) updateAll = true
       } else {
         index++
       }
@@ -2629,7 +2632,7 @@ module.exports = class Replicator {
 
     index = this._ranges.length === 0 ? 0 : index % this._ranges.length
 
-    return { checked, resolved, index }
+    return { checked, resolved, index, updateAll }
   }
 
   // "slow" updates here - async but not allowed to ever throw
@@ -2645,6 +2648,8 @@ module.exports = class Replicator {
         const result = this._updateRanges(rangeIndex, limit)
         const { checked, resolved } = result
         rangeIndex = result.index
+
+        if (result.updateAll) updateAll = true
 
         if (resolved > 0) checkedSinceResolve = 0
         else checkedSinceResolve += checked
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -2600,26 +2600,63 @@ module.exports = class Replicator {
     }
   }
 
+  _updateRanges(index, limit) {
+    if (this._ranges.length === 0) {
+      return { checked: 0, resolved: 0, hitMaxAfterResolve: false, index: 0 }
+    }
+
+    index = Math.min(index, this._ranges.length - 1)
+
+    let checked = 0
+    let resolved = 0
+    let hitMaxAfterResolve = false
+    let remaining = Math.min(limit, this._ranges.length)
+
+    while (remaining-- > 0 && this._ranges.length > 0) {
+      if (index >= this._ranges.length) index = 0
+
+      const r = this._ranges[index]
+
+      clampRange(this.core, r)
+      checked++
+
+      if (r.end !== -1 && r.start >= r.end) {
+        this._resolveRangeRequest(r)
+        resolved++
+        if (this._ranges.length === MAX_RANGES) hitMaxAfterResolve = true
+      } else {
+        index++
+      }
+    }
+
+    index = this._ranges.length === 0 ? 0 : index % this._ranges.length
+
+    return { checked, resolved, hitMaxAfterResolve, index }
+  }
+
   // "slow" updates here - async but not allowed to ever throw
   async _updateNonPrimary(updateAll) {
     // Check if running, if so skip it and the running one will issue another update for us (debounce)
     while (++this._updatesPending === 1) {
-      let len =
-        this._inflight.idle || updateAll
-          ? this._ranges.length
-          : Math.min(MAX_RANGES, this._ranges.length)
+      let checkedSinceResolve = 0
+      let rangeIndex = 0
 
-      for (let i = 0; i < len; i++) {
-        const r = this._ranges[i]
+      while (this._ranges.length > 0) {
+        const drain = this._inflight.idle || updateAll
+        const limit = Math.min(MAX_RANGES, this._ranges.length - checkedSinceResolve)
+        const result = this._updateRanges(rangeIndex, limit)
+        const { checked, resolved, hitMaxAfterResolve } = result
+        rangeIndex = result.index
 
-        clampRange(this.core, r)
+        if (hitMaxAfterResolve) updateAll = true
+        if (resolved > 0) checkedSinceResolve = 0
+        else checkedSinceResolve += checked
 
-        if (r.end !== -1 && r.start >= r.end) {
-          this._resolveRangeRequest(r)
-          i--
-          if (len > this._ranges.length) len--
-          if (this._ranges.length === MAX_RANGES) updateAll = true
-        }
+        if (checked === 0 || checkedSinceResolve >= this._ranges.length) break
+        if (!drain) break
+
+        await yieldToLoop()
+        if (this.destroyed || this._updatesPending > 1) break
       }
 
       for (let i = 0; i < this._seeks.length; i++) {
@@ -3386,6 +3423,10 @@ function incrementRx(stats1, stats2) {
 }
 
 function noop() {}
+
+function yieldToLoop() {
+  return new Promise((resolve) => setTimeout(resolve, 0))
+}
 
 function backoff(times) {
   const sleep = times < 2 ? 200 : times < 5 ? 500 : times < 40 ? 1000 : 5000
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -2015,6 +2015,8 @@ module.exports = class Replicator {
     this._active = 0
     this._ifAvailable = 0
     this._updatesPending = 0
+    this._updatesRunning = false
+    this._updatesQueued = false
     this._applyingReorg = null
     this._manifestPeer = null
     this._notDownloadingLinger = notDownloadingLinger
@@ -2637,60 +2639,78 @@ module.exports = class Replicator {
 
   // "slow" updates here - async but not allowed to ever throw
   async _updateNonPrimary(updateAll) {
-    // Check if running, if so skip it and the running one will issue another update for us (debounce)
-    while (++this._updatesPending === 1) {
+    this._updatesQueued = true
+    if (this._updatesRunning) {
+      if (this._inflight.idle || updateAll) this.queueUpdateAll()
+      return
+    }
+
+    this._updatesRunning = true
+
+    while (this._updatesQueued) {
+      this._updatesQueued = false
+
       let checkedSinceResolve = 0
       let rangeIndex = 0
       const drain = this._inflight.idle || updateAll
 
-      while (this._ranges.length > 0) {
-        const limit = Math.min(MAX_RANGES, this._ranges.length - checkedSinceResolve)
-        const result = this._updateRanges(rangeIndex, limit)
-        const { checked, resolved } = result
-        rangeIndex = result.index
+      while (true) {
+        this._updatesPending = 1
 
-        if (result.updateAll) updateAll = true
+        let continueRanges = false
+        if (this._ranges.length > 0) {
+          const limit = Math.min(MAX_RANGES, this._ranges.length - checkedSinceResolve)
+          const result = this._updateRanges(rangeIndex, limit)
+          const { checked, resolved } = result
+          rangeIndex = result.index
 
-        if (resolved > 0) checkedSinceResolve = 0
-        else checkedSinceResolve += checked
+          if (result.updateAll) updateAll = true
 
-        if (checked === 0 || checkedSinceResolve >= this._ranges.length) break
-        if (!drain) break
+          if (resolved > 0) checkedSinceResolve = 0
+          else checkedSinceResolve += checked
+
+          continueRanges = checked > 0 && checkedSinceResolve < this._ranges.length && drain
+        }
+
+        for (let i = 0; i < this._seeks.length; i++) {
+          const s = this._seeks[i]
+
+          let err = null
+          let res = null
+
+          try {
+            res = await s.seeker.update()
+          } catch (error) {
+            err = error
+          }
+
+          if (!res && !err) continue
+
+          if (i < this._seeks.length - 1) this._seeks[i] = this._seeks.pop()
+          else this._seeks.pop()
+
+          i--
+
+          if (err) s.reject(err)
+          else s.resolve(res)
+        }
+
+        this._updatesPending = 0
+
+        if ((!continueRanges || this._seeks.length > 0) && (this._inflight.idle || updateAll)) {
+          this.queueUpdateAll()
+        }
+        if (!continueRanges) break
 
         await yieldToLoop()
         if (this.destroyed) break
       }
 
-      for (let i = 0; i < this._seeks.length; i++) {
-        const s = this._seeks[i]
-
-        let err = null
-        let res = null
-
-        try {
-          res = await s.seeker.update()
-        } catch (error) {
-          err = error
-        }
-
-        if (!res && !err) continue
-
-        if (i < this._seeks.length - 1) this._seeks[i] = this._seeks.pop()
-        else this._seeks.pop()
-
-        i--
-
-        if (err) s.reject(err)
-        else s.resolve(res)
-      }
-
-      // No additional updates scheduled - break
-      if (--this._updatesPending === 0) break
-      // Debounce the additional updates - continue
-      this._updatesPending = 0
+      if (this.destroyed) break
     }
 
-    if (this._inflight.idle || updateAll) this.queueUpdateAll()
+    this._updatesPending = 0
+    this._updatesRunning = false
   }
 
   _clearRequest(peer, req) {
@@ -3427,7 +3447,7 @@ function incrementRx(stats1, stats2) {
 function noop() {}
 
 function yieldToLoop() {
-  return new Promise((resolve) => setTimeout(resolve, 0))
+  return new Promise((resolve) => setImmediate(resolve))
 }
 
 function backoff(times) {
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -2604,7 +2604,10 @@ module.exports = class Replicator {
   async _updateNonPrimary(updateAll) {
     // Check if running, if so skip it and the running one will issue another update for us (debounce)
     while (++this._updatesPending === 1) {
-      let len = Math.min(MAX_RANGES, this._ranges.length)
+      let len =
+        this._inflight.idle || updateAll
+          ? this._ranges.length
+          : Math.min(MAX_RANGES, this._ranges.length)
 
       for (let i = 0; i < len; i++) {
         const r = this._ranges[i]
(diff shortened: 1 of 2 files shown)


lib/replicator.js
```diff
@@ -2600,23 +2600,60 @@ module.exports = class Replicator {
...
 
-        clampRange(this.core, r)
+        if (resolved > 0) checkedSinceResolve = 0
+        else checkedSinceResolve += checked
 
-        if (r.end !== -1 && r.start >= r.end) {
-          this._resolveRangeRequest(r)
-          i--
-          if (len > this._ranges.length) len--
-          if (this._ranges.length === MAX_RANGES) updateAll = true
```
marcus-pousette-hp, line 2632: Removed the old this._ranges.length === MAX_RANGES special case because it only re-triggered updateAll when the queue shrank to exactly the cap, and I couldn’t find test coverage or PR discussion showing that boundary behavior is intentional.

marcus-pousette-hp: The actual MAX_RANGES cap is still preserved. The new cursor loop drains completed ranges in capped chunks without relying on that exact effect.

lib/replicator.js
```diff
@@ -2600,23 +2600,60 @@ module.exports = class Replicator {
...
-        }
+        if (checked === 0 || checkedSinceResolve >= this._ranges.length) break
+        if (!drain) break
+
+        await yieldToLoop()
```
mafintosh: why is this here?

marcus-pousette-hp: Previous behavior had implied event loop starvation prevention mechanics: 

https://github.com/holepunchto/hypercore/blob/c2ee97eac2b7047e4a5ca7a12a2180e0013dfd50/lib/replicator.js#L2607-L2609


In this PR I am trying to preserve this behavior by simultaniously also fixing the bug: 

```ts
while (this._ranges.length > 0) {
  this._updateRanges(...MAX_RANGES...)
  ...
  await yieldToLoop()
}
```

Also possible that debounce/coalescing mechanisms could be more probable to be invoked in `_updateNonPrimary` because of the yieldToLoop

marcus-pousette-hp: An additional improvement we can do here is to assert this behavior with a test

lib/replicator.js
```diff
@@ -3384,6 +3421,10 @@ function incrementRx(stats1, stats2) {
 
 function noop() {}
 
+function yieldToLoop() {
+  return new Promise((resolve) => setTimeout(resolve, 0))
```
```suggestion
  return new Promise((resolve) => setImmediate(resolve))
```
`setTimeout` defaults to a delay of `1` though the exact timing can be variable I believe.

lib/replicator.js
lejeunerenard: @marcus-pousette-hp the debouncing/guard via `_updatesPending` could delay seeks from being processed assuming a large number of ranges. This prevents from seeks being added here: https://github.com/holepunchto/hypercore/blob/ddd6c79eb5d545401a380cbca04c708b966ebbdf/lib/replicator.js#L1545

Here's a test case for asserting that seeks are processed relatively quickly. Before the PR it passes and after this PR it fails:
```js
test.solo('processing ranges doesnt block seeks', async function (t) {
  const core = await create(t)
  const clone = await create(t, core.key)

  // Well above MAX_RANGES so draining the backlog takes many batches: 50k/64
  const totalLength = 50_000

  await core.append(new Array(totalLength).fill('a'))

  // Make all range requests stay in `_ranges` forever so `_updateNonPrimary` has an oversized backlog to scan.
  await core.clear(0, totalLength)

  replicate(core, clone, t)

  // Populate ranges
  for (let i = 0; i < totalLength; i++) {
    clone.download({ start: i, end: i + 1 }).done().catch(noop)
  }

  t.is(clone.core.replicator._ranges.length, totalLength, 'large range backlog is pending')

  const bytesBefore = core.byteLength

  // Appending triggers an upgrade on clone, kicking off `_updateNonPrimary` to attempt the backlog.
  // Seeking to `bytesBefore` needs that same upgrade (so clone doesnt know about it).
  await core.append('end')

  // Baseline (no range backlog) resolves this in ~1ms; 900ms+ with the backlog
  const BUDGET = 100

  const start = Date.now()
  let delta = -1
  const seek = clone.seek(bytesBefore).then(() => {
    delta = Date.now() - start
  })
  await once(clone, 'append')

  // To allow merkle IO for seek
  await new Promise((resolve) => setImmediate(resolve))

  t.is(clone.core.replicator._seeks.length, 0, 'seek resolved after upgrade received')
  await seek

  t.ok(delta <= BUDGET, `seek resolved within ${BUDGET}ms despite a ${totalLength}-range backlog`)
  t.comment('seek time', delta)
  t.is(clone.core.replicator._ranges.length, totalLength, 'range backlog is still pending')
})
```

Test made with a litte AI assistance.

lib/replicator.js
lejeunerenard: I dont think this should be removed. The updateAll prompts requesting ranges from a random subset of peers potentially. Since the `_updateNonPrimary` function is called only when either receiving data or an upgrade, it could deadlock with requests queue in `_ranges` that haven't been requested of peers that can fulfill them if `updateAll()` isn't triggered.

The reason for the exact match is that we know in this scenario only 1 range was resolved so this check triggers as soon as there can be at least one range that could not have been requested of peers. Aka triggering another set of range requests could prompt a response which resolves a range not included in the last `_updateNonPrimary` run.

This doesn't prevent from large backlogs not getting stuck (hence the need for this PR) when there is more unfulfilled range request than the MAX_RANGES though. I don't think it should be `updateAll = true` for `this._ranges.length >= MAX_RANGES` though.

lejeunerenard (commented):
This approach is more thorough when completing range requests, but it blocks seek requests when processing large amounts of ranges. Accounting for this seems like it would add more state at the replicator level for tracking where the cursor left off etc.

If we keep `_updateNonPrimary`'s workload small but:
1. make it pick random subset of ranges to avoid stacking unfulfillable ranges at the front.
2. trigger non-primary updates more often, maybe when inflight goes idle

We can allow it to yield to seeks while being a minor change. I have the following competing PR with the above, which still doesn't pass atm. https://github.com/holepunchto/hypercore/pull/841

lib/replicator.js
marcus-pousette-hp: Ye, that make sense.  I restored this in https://github.com/holepunchto/hypercore/commit/5403eb5e7480440655351362b5fbc33e102d04f8, preserving the exact this._ranges.length === MAX_RANGES boundary. I also added a regression test with 65 one-block ranges and two peers as an proof/argumetn for this. Though it is quite verbose right now, I will re-iterate on this tomorrow.

marcus-pousette-hp:
One note, processing can get slow if there are many seeks at once or a very large backlog of ranges. Potentially  optimization/perf improrvement can be made here but left it out to make the PR easier

lib/replicator.js
```diff
@@ -2614,55 +2620,115 @@ module.exports = class Replicator {
...
-        const s = this._seeks[i]
+      while (true) {
+        const limit = Math.min(MAX_RANGES, this._ranges.length - checkedSinceResolve)
+        const result = this._updateRanges(rangeIndex, limit)
+        const { checked, resolved } = result
+        rangeIndex = result.index
 
-        let err = null
-        let res = null
+        if (result.updateAll) updateAll = true
```
```suggestion
        const { checked, resolved, index, updateAll: resultUpdateAll } = this._updateRanges(rangeIndex, limit)
        rangeIndex = index

        if (resultUpdateAll) updateAll = true
```

lib/replicator.js
```diff
@@ -2614,55 +2620,115 @@ module.exports = class Replicator {
...
 
-        if (!res && !err) continue
+          const seekIndex = this._seeks.indexOf(s)
+          if (seekIndex === -1 || (!res && !err)) continue
 
-        if (i < this._seeks.length - 1) this._seeks[i] = this._seeks.pop()
-        else this._seeks.pop()
+          if (seekIndex < this._seeks.length - 1) {
+            this._seeks[seekIndex] = this._seeks.pop()
+          } else this._seeks.pop()
```
lejeunerenard: This can be slightly simpler if we always pop and only swap if the popped version isn't the `s`. Something like this:
https://github.com/holepunchto/hypercore/blob/3cd8c50f3f38b4c20f63edd3c113cb91c4d4fef5/lib/replicator.js#L283-L286

lib/replicator.js
```diff
@@ -2614,55 +2620,115 @@ module.exports = class Replicator {
...
-        }
+        if (resolved > 0) checkedSinceResolve = 0
+        else checkedSinceResolve += checked
+
+        const continueRanges = checked > 0 && checkedSinceResolve < this._ranges.length && drain
+
+        this._updatingSeeks = true
+
+        let checkedSeek = false
+        for (const s of this._seeks.slice()) {
```
lejeunerenard, line 2692: Just a note as I dont think it is an issue in practice (because seeks are mainly used for streaming), but there is no limit to the number of seeks processed. This means just like ranges held up seek resolution, the opposite is still true where a large number of seeks could hold up range updates (allowing the initial update of a batch of ranges).

I propose we limit the length as well and use `RandomIterator`. Here is a test I made to showcase the issue:

```js
test('seeks cant block range updates', async function (t) {
  const a = await create(t)
  const b = await create(t, a.key)
  const totalBlocks = 1e5
  const totalRanges = 129 // above max range for at least 2x batches
  const totalSeeks = totalBlocks

  // Seed data
  await a.append(Array(totalBlocks).fill('a'))
  await a.clear(0, a.length - totalRanges) // Clear all but the end

  replicate(a, b, t)

  // Setup ranges at the end
  const ranges = []
  for (let i = 0; i < totalRanges; i++) {
    ranges.push(b.download({ start: a.length - (i + 1), end: a.length - i }).done())
  }

  // Setup large number of seeks
  const seeks = []
  for (let i = 0; i < totalSeeks; i++) {
    seeks.push(b.seek(i).catch(noop))
  }

  // Allow seeks to be added to the replicator
  await eventFlush()

  const result = await Promise.race([
    b.core.replicator._updateNonPrimary(true).then(() => 'resolved'),
    new Promise((resolve) => setTimeout(resolve, 100)).then(() => 'timeout')
  ])
  t.is(result, 'resolved', 'updating non-primary didnt block on seeks')
})
```

test/replicate.js
```diff
@@ -2937,6 +2937,348 @@ test('delayed updateAll timer doesnt keep event loop alive', async function (t)
...
+  await writer.append(new Array(totalLength).fill('a'))
+
+  const firstPeer = await create(t, writer.key)
+  let streams = replicate(writer, firstPeer, t, { teardown: false })
+  await firstPeer.get(0)
+  await unreplicate(streams)
+
+  const cappedPeer = await create(t, writer.key)
+  streams = replicate(writer, cappedPeer, t, { teardown: false })
+  await cappedPeer.get(totalLength - 1)
```
lejeunerenard: You don't need to disable the `teardown` for these `replicate()` calls. You can `unreplicate` with it in place.

test/replicate.js
```diff
@@ -2937,6 +2937,348 @@ test('delayed updateAll timer doesnt keep event loop alive', async function (t)
   t.is(r._updateAllBump, null, 'timer reset to null')
 })
 
+test('completed range exposes capped range to existing peer', async function (t) {
```
lejeunerenard: Not sure what the purpose of this test is besides that peers can get disparate ranges from different peers. It passes on `main` so isn't part of the idle range fix. Maybe it is trying to capture that there is a `MAX_RANGES` batch size per `_updateNonPrimary()` but it doesn't assert anything about the `_updateNonPrimary()` lifecycle.

test/replicate.js
```diff
@@ -2937,6 +2937,348 @@ test('delayed updateAll timer doesnt keep event loop alive', async function (t)
...
+  let seek = null
+
+  replicator._updateRanges = function (index, limit) {
+    const result = updateRanges.call(this, index, limit)
+    checked += result.checked
+
+    if (++updates === 1) {
+      setImmediate(() => {
+        peer.paused = false
+        seek = clone.seek(Math.floor(core.byteLength / 2))
```
lejeunerenard: I dont understand why we need to wait to process a single batch of range requests to schedule the seek request for after the poll phase in the loop. Calling here `let seek = clone.seek(...)` works too.

lejeunerenard (commented):
Did another pass. Still need to cover some of the tests but found a potential way that ranges now can be blocked instead of seeks.

test/replicate.js
marcus-pousette-hp: Fixed ! https://github.com/holepunchto/hypercore/pull/832/commits/176da453c02f39ad093213c8da9d68c8752e6951

lib/replicator.js
marcus-pousette-hp: Good idea! Fixed https://github.com/holepunchto/hypercore/pull/832/commits/176da453c02f39ad093213c8da9d68c8752e6951

lib/replicator.js
marcus-pousette-hp: Fixed https://github.com/holepunchto/hypercore/pull/832/commits/176da453c02f39ad093213c8da9d68c8752e6951

lib/replicator.js
marcus-pousette-hp: Yup but this issue also seem to be present main. But I would argue here still that RandomIterator is not a good enough because skipped seeks somehow would need a gauranteed pass acrsos additions/cancellations. I would prefer to address this in a future PR, perhaps using an alternating approach between RangeRequest and SeekRequest processing, tuned around the trade-offs of which requests should be prioritized and how much work each batch should perform.

marcus-pousette-hp: Also, for that follow up, perhaps we can define a regression case or performance metric that demonstrates the benefit of RandomIterator probabilistic selction

test/replicate.js
marcus-pousette-hp: Yep, this adds coverage for the invariant  https://github.com/holepunchto/hypercore/pull/832#discussion_r3554921627. 
The test passes on main, as expected, because the behavior already exists there. 

Before adding this test, I could remove the boundary behavior without any existing test failing, which felt wrong so I added it here.

The lifecycle part is intentional: the test asserts observable behavior rather than counting _updateNonPrimary() calls directly. Since the normal idle rescan is disabled (through the inflight mutation we do in the test), the final download only succeeds if crossing from 65 to 64 causes existing peers to be rescanned.

test/replicate.js
marcus-pousette-hp: You’re right, setImmediate() isn’t needed. clone.seek() yields during its initial local lookup, so _updateNonPrimary() processes the first 64-range batch before the seek is added and requested. Simplified in 64bc2a5.

lib/replicator.js
lejeunerenard: You're right. Lets do it in a separate PR. Also noticed you mentioned this issue last week:
https://github.com/holepunchto/hypercore/pull/832#issuecomment-4990557588
Too many comments on this PR, further evidence that it should be broken out to a separate PR.

test/replicate.js
lejeunerenard: Ah! Sorry about that. Makes sense now.

lib/replicator.js
```diff
@@ -2614,55 +2620,118 @@ module.exports = class Replicator {
...
+
+        this._updatingSeeks = false
+
+        if (
+          (!continueRanges || (checkedSeek && this._seeks.length > 0)) &&
+          (this._inflight.idle || updateAll)
+        ) {
+          this.queueUpdateAll()
+        }
+        if (!continueRanges) break
```
lejeunerenard, line 2724: Mostly a note for the future PR to rework seeks as well, but these two conditions probably can be combined. It seems to me if we are updating peers, we likely need to break the loop.

Previously this wasn't a question since it was at the end of the non-primary update anyways. Now with it looping more, it means that we could request more range blocks and/or seeks. Ranges can break the inner loop but seeks cannot. So currently the loop is biased toward ranges (makes sense for the scope of a fix for idle range).

FYI this change would require updating `range drain checks unresolved seeks once` ~and likely removing the `drain` check above as that's currently only for whether the range requests should continue. If we assume that the signal for `drain`ing is enough in combination with either ranges or seeks, then no need to check ahead~.

lib/replicator.js
```diff
@@ -2614,55 +2620,118 @@ module.exports = class Replicator {
...
 
-      // No additional updates scheduled - break
-      if (--this._updatesPending === 0) break
-      // Debounce the additional updates - continue
-      this._updatesPending = 0
+      if (this.destroyed) break
     }
 
-    if (this._inflight.idle || updateAll) this.queueUpdateAll()
+    this._updatingSeeks = false
```
lejeunerenard: nit: I believe this is unnecessary right? `_updatingSeeks` is only modified in the loop above and there is no way currently for it to be left as `true`.

test/replicate.js
```diff
@@ -2937,6 +2937,342 @@ test('delayed updateAll timer doesnt keep event loop alive', async function (t)
...
+  const updateRanges = replicator._updateRanges
+  let updates = 0
+
+  replicator._updateRanges = function (index, limit) {
+    const result = updateRanges.call(this, index, limit)
+
+    if (++updates === 1) {
+      setImmediate(() => {
+        this._inflight._active++
+        this._updateNonPrimary(false).catch(noop)
```
lejeunerenard: I'm generally not sure about what this test is asserting. But this `_updateNonPrimary(false)` will always hit the short circuit return at the beginning of the function because `_updatesRunning` will always be true. It wont queue the updateAll either since the inflight is manually set to be active & the `updateAll` arg is false.

lejeunerenard (commented):
Only one test i'm not sure what it's purpose is, but otherwise only notes about seeks vs ranges behavior for the PR in the future.

test/replicate.js
lejeunerenard: I get the test assertion now, but the `_updateNonPrimary(false)` doesn't exercise any code paths still. I assume this is to emulate how either a data message or an upgrade might have retriggered the `_updateNonPrimary()` call.

Probably can be removed given the `inflight._active++` is the main thing being tested and shows that something is inflight which normally would interrupt the drain without the `drain` variable storing the original intent to drain.

lib/replicator.js
lejeunerenard: Per this comment https://github.com/holepunchto/hypercore/pull/832#discussion_r3633325052 I now understand the purpose of the new `drain` variable storing a value. It's still a question of what that will look like for a future seeks PR.

lib/replicator.js
marcus-pousette-hp: Yea, fixed!

test/replicate.js
marcus-pousette-hp: _updateNonPrimary(false) does return early, but first sets _updatesQueued = true. That is separate from _inflight._active++: the test covers both new inflight work and another update being queued during the yield. Removing the call would lose coverage of the queued-update race. I clarified this in an inline comment  https://github.com/holepunchto/hypercore/pull/832/commits/2407590adf963a661de44c22d9d34dcdbd0ef1e6

lib/replicator.js
marcus-pousette-hp: Yes I agree

test/replicate.js
lejeunerenard: Didn't think about `_updatesQueued` being set, but it doesn't matter in this case because the `drain` keeps the inner loop going. The `_updatesQueued` makes the outer loop go another time, but it does no work because it isnt set to drain by then.

If we want to cover this additional loop in the test case, then we should assert the value of `updates` at the end. Naively I thought this would be the number of batches from one pass (`Math.ceil(totalLength / 64)`) plus the additional pass, but I didn't think about how the `checkedSinceResolve` is updated once a range is resolved. So the actual assertion looks like the following:
```js
  const singlePass = Math.ceil(totalLength / 64)
  const repassBecauseOfResolving = Math.ceil(availableStart / 64)
  const firstPassBatches = singlePass + repassBecauseOfResolving
  t.is(updates, firstPassBatches + 1, 'does a second pass because _updateNonPrimary called mid update')
```
Does point out that the work done during a `drain` is not optimal but since this covers the case where ranges are added, I think it's fine as is. A more efficient approach can be part of refactoring seeks as well.

Also as an aside, the `.catch(noop)` isn't necessary right? `_updateNonPrimary` isn't allowed to throw.

marcus-pousette-hp: My gut feeling says no about adding extra assertion here for now, since what we are testing that the queued update must not interrupt the current drain, and the extra assertion propose would make it overly strict towards future changes. I did the change fixing the noop catch though.

lejeunerenard: Ok then there is no purpose to the `_updateNonPrimary` call correct? It causes the outer loop to trigger again (by setting `_updatesQueued`) but the assertions don't cover that without the above addition. Without an assertion, it should be removed as well as the `eventFlush()` (and `inflight._active = 0` could be too).

I could see either way as fine, as they essentially one of two scenarios:
1. Testing that `drain` fully drains even when inflight blocks are sent when yielding the loop.
2. The same as above, with the additional coverage of another update queued.

lejeunerenard: Opted to keep the `_updateNonPrimary` in the test to assert how it queues an runs after a full drain. Change here: 7e81e3a

lejeunerenard (approved):

PR #840, opened 2026-07-07
enable TreeNode cache (proof cache)

telamon:
dep:  https://github.com/holepunchto/hypercore-storage/pull/124


mafintosh:
Approved, but we should bump the min storage version in the semver also before it lands for explicitly clarity (for future us when looking at the commit)

mafintosh (dismissed):

lejeunerenard:
Requires https://github.com/holepunchto/hypercore-storage/pull/125 for min `hypercore-storage` version

PR #851, opened 2026-07-20
add suspend controller for replication

tur-nr:
This adds a way to pause replication without dropping connections. `SuspendController` works like an `AbortController`: you hold the controller and pass `controller.signal` into cores via the new `suspendSignal` option. One signal can be shared by any number of cores, and a core opened while the signal is suspended comes up suspended.

While suspended, a core sends no requests, uploads no blocks and drops incoming data (clearing the request so it gets re-fetched later). Incoming requests are queued in the existing receiver queue and answered on resume, and the replicator listens for the signal's resume event to serve that backlog and re-request whatever was dropped. Wants are still answered since they're served from memory. `replicator._replicationBusy()` lets the caller wait until in-flight serves/verifies have finished after suspending.

The reason we need this: consumers can briefly suspend hypercore to release resources and locks, for example suspending the underlying storage, whilst keeping the swarm and peers warm. Without it, any replication traffic against suspended storage parks forever, so the only safe option today is tearing connections down first. With this, storage can suspend underneath live connections and everything picks up where it left off on resume, no reconnect needed.

Cores that don't pass a signal are completely unaffected.

### DIFF
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -2124,8 +2124,9 @@ module.exports = class Replicator {
         peer.dataProcessing > 0 ||
         peer.syncsProcessing > 0 ||
         peer.pushProcessing > 0
-      )
+      ) {
         return true
+      }
     }
     return false
   }
--- a/lib/replicator.js
+++ b/lib/replicator.js
@@ -885,6 +885,8 @@ class Peer {
   // Safe to call in the background - never fails
   async _canUpgrade(remoteLength, remoteFork) {
     if (remoteFork !== this.core.state.fork) return false
+    // upgradeable reads the merkle tree, no storage reads while suspended
+    if (this.replicator.suspended) return false
 
     if (remoteLength === 0) return true
     if (remoteLength >= this.core.state.length) return false
@@ -942,7 +944,7 @@ class Peer {
       this.remoteRequests.set(msg.id, msg)
     }
 
-    if (!this.protomux.drained || this.receiverQueue.length) {
+    if (!this.protomux.drained || this.receiverQueue.length || this.replicator.suspended) {
       this.receiverQueue.push(msg)
       return
     }
@@ -966,7 +968,7 @@ class Peer {
   }
 
   async _handleRequests() {
-    if (this.receiverBusy || this.replicator.destroyed) return
+    if (this.receiverBusy || this.replicator.destroyed || this.replicator.suspended) return
     this.receiverBusy = true
     this.protomux.cork()
 
@@ -974,7 +976,8 @@ class Peer {
       this.remoteOpened &&
       this.protomux.drained &&
       this.receiverQueue.length > 0 &&
-      !this.removed
+      !this.removed &&
+      !this.replicator.suspended
     ) {
       const msg = this.receiverQueue.shift()
       await this._handleRequest(msg)
@@ -985,6 +988,7 @@ class Peer {
   }
 
   async push(index) {
+    if (this.replicator.suspended) return
     if (!this.remoteAllowPush) return
     if (!this.remoteDownloading) return
     if (this.core.state.length <= index) return
@@ -1204,6 +1208,16 @@ class Peer {
   }
 
   async _handleData(data) {
+    // drop data while suspended, clearing the request so resume re-requests it
+    if (this.replicator.suspended) {
+      const req = data.request > 0 ? this.replicator._inflight.get(data.request) : null
+      if (req !== null && req.peer === this) {
+        this._onrequestroundtrip(req)
+        this.replicator._onnodata(this, req, 0)
+      }
+      return
+    }
+
     // always allow a fork conflict proof to be sent
     if (data.request === 0 && data.upgrade && data.upgrade.start === 0) {
       if (await this.core.checkConflict(data, this)) return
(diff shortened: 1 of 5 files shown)


test/suspend-replication.js
```diff
@@ -0,0 +1,112 @@
...
+  await sleep(200)
+
+  t.is(b.length, 3, 'no upgrade while suspended')
+
+  controller.resume()
+
+  t.alike(await b.get(4), Buffer.from('e'), 'caught up after resume')
+})
+
+test('suspended replication queues incoming requests, serves them on resume', async function (t) {
```
lejeunerenard, line 33: This test doesn't have assertions for the fact that it queues messages. When running it with a log in `_handleRequests` it doesn't ever log that it is processing the queue.

lib/replicator.js
lejeunerenard, line 1115: The fact that `isActive()` is based on `suspended` means that fulfilling request can bail here leaving requests not fulfilled. This can deadlock in the following test:

```js
test('suspended replication queues incoming requests, doesnt throw when suspending microtick after resume', async function (t) {
  const controller = new Hypercore.SuspendController()

  const a = await create(t, null, { suspendSignal: controller.signal })
  const b = await create(t, a.key)
  await a.append(['a', 'b', 'c'])

  const bAppended = once(b, 'append')
  replicate(a, b, t)
  await bAppended // ensure b has latest length

  controller.suspend()
  const get = b.get(0)
  const early = await Promise.race([get.then(() => 'served'), sleep(300).then(() => 'pending')])
  t.is(early, 'pending', 'request not served while suspended')

  controller.resume()
  await eventFlush()
  controller.suspend()

  await eventFlush()
  controller.resume()

  t.alike(await get, Buffer.from('a'), 'queued request served after resume')
})
```

Note you need to mine to hit the scenario.

So when handling queued requests the `while` checks if the replicator is suspended, but it can become suspended mid-`_handleRequest()`.

lejeunerenard (changes requested):
While reviewing I found a few fixes and since they were simple, I just pushed them. There were:

- `replicator.busy` didn't reflect push mode requests which also use storage.
- Possible deadlock if throwing an error while handling a request after resuming suspension.

Also pointed out how there is still an issue around guarding against suspending while handling requests that were queued.

Reviewed with help from AI.

lib/replicator.js
tur-nr: Fixed in 6df919f660865841bbafb51da743ceb1cff37509

test/suspend-replication.js
tur-nr: Fixed in b09fd45f7806d84b99cdf199c92935f1e73ef235

_*AI used._

lib/replicator.js
```diff
@@ -1098,6 +1107,15 @@ class Peer {
...
       this.remoteRequests.delete(req.msg.id)
+
+      // suspended mid-flight, after async req.fulfill(). This request was
+      // already unregistered, so falling through to the isActive() below
+      // would drop it. Putting it back for the next _handleRequests() call.
+      if (this.replicator.suspended && !this.removed && proof.block !== null) {
+        this.remoteRequests.set(req.msg.id, req.msg)
+        this.receiverQueue.push(req.msg)
+        return
+      }
```
lejeunerenard: This fix requeues requests already completed to avoid the `isActive()` check below (mentioned here: https://github.com/holepunchto/hypercore/pull/851/changes#r3617257798). This is the wrong approach because the issue is not that the request shouldn't be fulfilled, but that a valid response is being dropped. The replicator can be suspended while its going to storage for fulfilling the request. Afterwards there is no more storage IO to block suspension, so we can send it correctly.

The issue is that `isActive()` is used here as a guard against 'bad behaving' peers rather than whether the replicator is active. I think this actually points out that using `isActive()` as a guard is misusing it as it is a per `Peer` method. We should add `replicator.suspend` instead to all places that `.isActive()` is used currently. The following line would need to be reverted too as it was excluding suspension as the reason to not reply to wants:
https://github.com/holepunchto/hypercore/pull/851/changes#diff-3344a84994d8e172768f33ec4bf514ba2bc28eb7eeffff10fcbbd294efe56edcR2863

There are other places that currently were skipped when suspended (such as broadcasting the range) that could be allowed when suspended since they dont use IO / storage, much like responding to wants.

If we do want to stop messages sending while suspended even after the IO is done, we can requeue the msg, but:
1. dont need to check `this.removed` as it cannot be true given we just checked that the remoteRequest for it. It would be cleared if removed.
2. Can't check `proof.block` as the fulfilled proof could be `null` etc.
3. Should be moved to before we remove the request from the `remoteRequests` right above.

lejeunerenard (commented):
Thanks for the fixes. If I understand the problem we're trying to solve, IO is the culprit for issues suspending, not networking traffic. This means we could keep more network sent while suspended. Then we should refactor so `isActive()` doesn't include the replicator being suspended as it was primarily used for 'bad behaving' peers with the exception of the core being frozen due to a conflict.

lib/replicator.js
tur-nr: Thanks for review again. Sorry I misunderstood what is active means so yeah putting the suspended check there isn't the solution.

Yes you are correct with network can continue and suspend only when idle.

I've reverted the `isActive` work and added distinct `.suspended` checks at points where new or upgraded blocks are request.

Fixed: 36c46f2ff7c54822f13922806d5a4dee4c2fd720

PR #852, opened 2026-07-23
support manifest as raw buffer

chm-diederichs:


test/manifest.js
```diff
@@ -1518,6 +1518,205 @@ test('manifest - persist if manifest is updated', async function (t) {
...
+        signature: 'ed25519',
+        publicKey: keyPair.publicKey
+      }
+    ]
+  })
+
+  const key = Verifier.manifestHash(manifest)
+  const create = await createStored(t)
+
+  const core = await create(key, { compat: false })
```
lejeunerenard, line 1550:
```suggestion
  const core = await create(key, { compat: false })
  t.teardown(() => core.close())
```

`createStored()` doesn't include the teardown for closing the core.

test/manifest.js
```diff
@@ -1518,6 +1518,205 @@ test('manifest - persist if manifest is updated', async function (t) {
...
```
lejeunerenard, line 1621:
```suggestion
  const core = await create(key, { compat: false })
  t.teardown(() => core.close())
```

lejeunerenard (commented):
Took a high level pass looking for why the tests are failing. Found a few places where the core isnt closed on teardown.

Let me know when I can look in more depth.

index.js
```diff
@@ -130,11 +132,31 @@ class Hypercore extends EventEmitter {
...
 
+  static parseManifest(manifest, key) {
+    const parsed = createManifest(manifest)
+    if (parsed === null) return null
+
+    if (parsed.quorum > parsed.signers.length) {
+      throw BAD_ARGUMENT('Quorum should not be higher than the number of signers')
+    }
+
+    new Verifier(manifestHash(parsed), parsed)
```
lejeunerenard: Is this line for verifying the parsed manifest? Using a class construction for validation is kinda yucky.

index.js
```diff
@@ -130,11 +132,31 @@ class Hypercore extends EventEmitter {
...
+          : this.parseManifest(manifest)
     }
     return compat ? manifest.signers[0].publicKey : manifestHash(createManifest(manifest))
   }
 
+  static parseManifest(manifest, key) {
+    const parsed = createManifest(manifest)
+    if (parsed === null) return null
+
+    if (parsed.quorum > parsed.signers.length) {
```
lejeunerenard, line 147: Feels like we should refactor validating manifests so rules like this one are not repeated (also in `lib/verifier.js`). Refactoring can be a future PR though.

index.js
mafintosh: Agree. Should be a static method if validation. Can you refactor?

lejeunerenard (approved):

PR #861, opened 2026-08-20
Skip errors when replicator destroyed

lejeunerenard:
This is a proposed refinement of #859. It opted to check if the core was closing to skip processing any messages assuming there was still network IO, but I believe this is not the case as the channels are closed when the replicator is destroyed and closed both which happen when closing the core.

Instead the issue is that the requests already came in and have pending disk IO during which the core and the replicator is closed. To guard against errors from closing the channel mid process, a new [Protomux PR](https://github.com/holepunchto/protomux/pull/31) which warns when a handler / callback throws instead catches the errors and reemits as `warning` events.

One exception to the new protomux handler catching is the case originally tested in #859 where the error comes from `_send()`. This used `this.stream.destroy(err)` directly to report the error. Now it is guarded to warn instead (via protomux's method) when the replicator is destroyed.

Depends on:

- https://github.com/holepunchto/protomux/pull/31


lejeunerenard:
Updated the existing test and added another for processing a message (`data`) which relies on protomux directly.

One catch to the protomux setup is that it emits `warning` which doesn't propagate through the NoiseStream normally setup by `core.replicate()`.  So using the normal `replicate()` helper function in the tests didn't work. .

lib/replicator.js
```diff
@@ -1960,8 +1957,9 @@ class Peer {
         }
       }
     } catch (err) {
-      if (this.core.closing !== null) return
-      this.stream.destroy(err)
+      if (this.removed) this.protomux._warn(err)
```
mafintosh: private api

lejeunerenard: Resolved in [d04289c](https://github.com/holepunchto/hypercore/pull/861/commits/d04289cd26f52652457b936772456cc223f4f5d1)

mafintosh (approved):

